PR #29, opened 2018-10-01
add flow starts

nicpottier:
depends on rapidpro pr which is incoming.


nicpottier:
also needs more testing.

nicpottier:
now with testing!

nicpottier:
merging because I want to push to staging but still want eyes on this.

PR #56, opened 2018-12-21
✨ Update to latest goflow

rowanseymour:


Gopkg.lock
```diff
@@ -2,158 +2,203 @@
 
 
 [[projects]]
+  digest = "1:5e79f7a65816b2ba311439df21c9b24af91868cbb5d67489be6d94be2cccb286"
```
rowanseymour, line 5: looks like newer `dep` adds some extra stuff

nicpottier:
This isn't a go module setup though is it? IE, you aren't using https://github.com/golang/go/wiki/Modules on the mailroom side but still dep right?

rowanseymour:
@nicpottier no this is leaving mailroom as a non-module project but being able to use goflow as a module.. was assuming that's what you wanted for now

Issue #60, opened 2019-01-10
Add Surveyor Support


nicpottier:
Allow surveyor results to be posted to mailroom endpoint
 * [x] set up test harness, submit basic fully valid submissions
 * [ ] deal with modifiers on groups / fields that no longer exit
 * [x] deal with modifiers that have no associated events (ie, removal from group that exists in db but wasn't in surveyor)

nicpottier:
So @rowanseymour remind me again our discussion around modifiers? One goal is to get the contact in the right state to pass to the hooks that process the events on a session (IE, merge the events with those that occurred) but I think another was also to bring up events which wouldn't have otherwise caused any action right?

The example I can think of is removal of the contact from a group (that the contact was never added to in surveyor). In that case, there is no event since we are recording state change, BUT, that modifier may generate an event if the DB contact was indeed in that group, correct? Just making sure I'm thinking of this right.

Only oddity is we do have the possibility of duplicating some of this work, in that we will run through the contact modifying events both with the modifiers that generate events but also when processing the normal flow events. Guess I could just filter those out though to prevent that since our contact is in the final state at that point anyways.

nicpottier:
So I took care of the modifier case.. I think the only missing piece to this being complete is being able to deal with events / sessions that  have references to fields / groups that no longer exist. We parse these one by one, so I *could* just throw away any that throw an error when being unmarshalled, but that seems a bit too coarse. If feels to me like the validation step either needs to be separate (first you unmarshal, then you validate?) or we need some kind of type error for "missing fields / groups" so that I can deal with those differently. IE, I don't want to silently throw away events that return an error when unmarshalling as maybe they are failing for more serious errors. (bad format, whatever else)

rowanseymour:
In theory everything in the session blob should be readable after assets have ceased to exist. Things like events and triggers only ever contain references to assets which are only resolved if/when needed at execution time.

Exceptions are `Contact` and `Input`. They might be need to be resolved in an expression and currently `Resolve` isn't aware of things like assets.. so they resolve their assets at load time. In the case of `Contact` being loaded with a session, there is an argument `strict` which is set to `false` and means we don't error if a field or group is missing.

Do agree that it would be nice if this was more explicitly different to other kids of loading errors.

And looking at the code now... it looks like we still might blow up in the following cases:

- loading `session.input.channel` if channel no longer exists
- loading `session.contact.urns` if channel no longer exists

[...]

PR #65, opened 2019-01-27
📋 allow submission of surveyor flows, add tests

nicpottier:
Still missing is how we deal with events / modifiers that fail validation due to groups no longer existing in the database, but otherwise complete.

Fixes: #60

### DIFF
--- a/web/surveyor/surveyor.go
+++ b/web/surveyor/surveyor.go
@@ -147,7 +147,7 @@ func handleSubmit(ctx context.Context, s *web.Server, r *http.Request) (interfac
 	}
 
 	// create our sprint
-	sprint := engine.NewSprintWithValues(contactModifiers, modifierEvents)
+	sprint := engine.NewSprint(contactModifiers, modifierEvents)
 
 	// write our session out
 	tx, err := s.DB.BeginTxx(ctx, nil)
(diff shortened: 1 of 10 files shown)


web/surveyor/surveyor.go
```diff
@@ -52,53 +64,130 @@ func handleSurveyorSubmit(ctx context.Context, s *web.Server, r *http.Request) (
...
 		}
 		sessionEvents = append(sessionEvents, event)
 	}
 
+	// and our modifiers
+	contactModifiers := make([]flows.Modifier, 0, len(request.Modifiers))
+	for _, m := range request.Modifiers {
+		modifier, err := modifiers.ReadModifier(assets, m)
+		if err != nil {
+			return nil, http.StatusBadRequest, errors.Wrapf(err, "error unmarshalling modifier: %s", string(m))
```
rowanseymour, line 99: to be clear, `ReadModifier` currently might fail for reasons besides unmarshalling - i.e. can't find an asset

rowanseymour (approved):

web/surveyor/surveyor.go
nicpottier: Yep, see PR comment. :)

PR #111, opened 2019-03-08
👐Add support to handle welcome message event as noop

norkans7:
In addition to 
https://github.com/nyaruka/courier/pull/222
https://github.com/nyaruka/rapidpro/pull/2347

to address https://github.com/rapidpro/rapidpro/issues/883

### DIFF
--- a/handler/worker.go
+++ b/handler/worker.go
@@ -335,12 +335,6 @@ func HandleChannelEvent(ctx context.Context, db *sqlx.DB, rp *redis.Pool, eventT
 		return nil, errors.Errorf("unknown channel event type: %s", eventType)
 	}
 
-	// no trigger, noop, move on
-	if trigger == nil {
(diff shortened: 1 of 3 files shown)


handler/worker.go
```diff
@@ -327,6 +328,9 @@ func HandleChannelEvent(ctx context.Context, db *sqlx.DB, rp *redis.Pool, eventT
 	case models.MOCallEventType:
 		trigger = models.FindMatchingMOCallTrigger(org, modelContact)
 
+	case models.WelcomeMessateEventType:
+		trigger = nil
+
```
nicpottier, line 334: So I think this is right, but we need to move the `trigger == nil` stuff below to actually be below where we update preferred URN and calculate dynamic groups. That way we know the new contact case will be dealt with.

nicpottier (changes requested):

PR #141, opened 2019-06-21
🚨 Fix endpoints so we don't hard error for expected requests

rowanseymour:
Ended up tweaking things to try and clarify the contract for handler functions. Sometimes you want to return an error as a normal response (e.g. a flow failed validation, goflow gives us an error describing why), and sometimes an error means something unexpected happened, please log to sentry.

Handlers have 3 return values: response, status_code, err. So this change means..

 - If you want your error to be a response, return it as the response with a status code
 - If you want your error to be a hard error, just return it as the error, and nothing for the other values

### DIFF
--- a/web/flow/flow.go
+++ b/web/flow/flow.go
@@ -31,19 +31,18 @@ func init() {
 //   }
 //
 type migrateRequest struct {
-	Flow      json.RawMessage `json:"flow"            validate:"required"`
-	IncludeUI *bool           `json:"include_ui"` // ignored
+	Flow json.RawMessage `json:"flow" validate:"required"`
 }
 
 func handleMigrate(ctx context.Context, s *web.Server, r *http.Request) (interface{}, int, error) {
 	request := &migrateRequest{}
 	if err := utils.UnmarshalAndValidateWithLimit(r.Body, request, web.MaxRequestBytes); err != nil {
-		return nil, http.StatusBadRequest, errors.Wrapf(err, "request failed validation")
+		return errors.Wrapf(err, "request failed validation"), http.StatusBadRequest, nil
 	}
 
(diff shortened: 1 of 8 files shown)


web/flow/flow.go
```diff
@@ -31,19 +31,18 @@ func init() {
 //   }
 //
 type migrateRequest struct {
-	Flow      json.RawMessage `json:"flow"            validate:"required"`
-	IncludeUI *bool           `json:"include_ui"` // ignored
```
rowanseymour, line 35: unrelated - this just doesn't need to be here

web/flow/flow.go
```diff
@@ -201,20 +201,20 @@ func readFlow(flowDef json.RawMessage) (flows.Flow, error) {
 	return flow, nil
 }
 
-func validate(ctx context.Context, db *sqlx.DB, orgID models.OrgID, flow flows.Flow) (int, error) {
+func validate(ctx context.Context, db *sqlx.DB, orgID models.OrgID, flow flows.Flow) (interface{}, int, error) {
 	org, err := models.NewOrgAssets(ctx, db, orgID, nil)
 	if err != nil {
-		return http.StatusBadRequest, err
+		return nil, 0, err
```
rowanseymour, line 207: This is not ideal... but we don't know if `models.NewOrgAssets` fails because orgID isn't a valid org (would expect that to be a client error), or because the database is down (would expect that to be a server error)

So erring on the side of harder error.

nicpottier (approved):
I like it! Probably shouldn't merge this until 5.1 though.

rowanseymour:
@nicpottier this good to go now?

nicpottier (approved):

PR #144, opened 2019-07-02
🔍   Starting contacts into flow via ElasticSearch

nicpottier:
Allows starting a set of contacts in a flow, our start job just needs a `query` parameter.

Will require ansible updates to update mailroom env to know about Elastic.


nicpottier:
@dodobas would appreciate a close review of this!

elastic/elastic.go
```diff
@@ -0,0 +1,330 @@
...
+				name = "fields.ward"
+			} else if field.Type == District {
+				name = "fields.district"
+			} else if field.Type == State {
+				name = "fields.state"
+			}
+			name += "_keyword"
+
+			if c.Comparator() == "=" {
+				query = q.NewTermQuery(name, value)
```
dodobas: i think we are missing `fieldQuery` for  `=` and `!=` operators, json test cases should also be updated

nicpottier: I think I am including the field but more importantly I wasn't building a nested query, updating.

nicpottier:
@dodobas updated, can you take another pass?

ivr/ivr_test.go
```diff
@@ -32,10 +32,10 @@ func TestIVR(t *testing.T) {
 	db.MustExec(`UPDATE channels_channel SET channel_type = 'ZZ', config = '{"max_concurrent_events": 1}' WHERE id = $1`, models.TwilioChannelID)
 
 	// create a flow start for cathy
-	start := models.NewFlowStart(models.Org1, models.IVRFlow, models.IVRFlowID, nil, []models.ContactID{models.CathyID}, nil, false, true, true, nil, nil)
+	start := models.NewFlowStart(models.Org1, models.IVRFlow, models.IVRFlowID, nil, []models.ContactID{models.CathyID}, nil, "", false, true, true, nil, nil)
```
nicpottier: This constructor is getting a bit out of hand, going to tweak it to do `With()` for recipients.

hooks/session_triggered.go
```diff
@@ -97,12 +97,13 @@ func (h *InsertStartHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool
...
-				true, true,
-				event.RunSummary, nil,
-			)
+			start := models.
+				NewFlowStart(org.OrgID(), flow.FlowType(), flow.ID(), models.DoRestartParticipants, models.DoIncludeActive).
+				WithGroupIDs(groupIDs).
+				WithContactIDs(contactIDs).
+				WithURNs(event.URNs).
+				WithCreateContact(event.CreateContact).
+				WithParentSummary(event.RunSummary)
```
nicpottier, line 105: golang doesn't support using `()` around a block like this the way Python does. Having a trailing `.` is the only way to get this not reformatted that I could figure out. Slightly weird but I bet we get used to it.

models/starts.go
```diff
@@ -18,6 +18,18 @@ type StartID null.Int
...
+type RestartParticipants bool
+
+const DoRestartParticipants = RestartParticipants(true)
+const DontRestartParticipants = RestartParticipants(false)
+
+// IncludeActive is our type for the bool of whether to include active contacts
+type IncludeActive bool
+
+const DoIncludeActive = IncludeActive(true)
+const DontIncludeActive = IncludeActive(false)
```
nicpottier, line 31: Death to bool parameters!

hooks/session_triggered.go
rowanseymour: been doing this in goflow for a while with its builder classes, except with `NewFlowStart(..)` on the same line as `models.` and only the `With` lines indented... which seems marginally less weird, maybe?

nicpottier:
So random thoughts on allowing starts via query..

One thing we have heard is that it would be nice to be able to start a flow for a group of contacts and exclude some contacts, specifically contacts which have a WhatsApp URN. (or maybe more precisely those that would be sent to using a WhatsApp URN, but lets ignore that specificity for this discussion)

`query` as I've added here is used as a union of the other fields, the same way that you can have both `contact_ids` and `group_ids` for a FlowStart and we will start the union of those contacts. (uniqued by `contact_id`)

I *think* that's right, but it does not allow for the scenario above.

Two ways we could accomplish that:
1. Have an `exclude_query` field that basically subtracts contacts from the set of contacts that will be started in a flow. Seems a bit weird but would be simple enough to add.
2. Change all our operations to turn into a query. IE, if you say you want to start Bob and all Reporters in a flow, then we turn that into `id=134 OR group=134-1234-1234-1234-1234` and we just set the query field. That would change to basically say that all starts now go through Elastic and are treated the same.

Option 2 above is appealing from a number of code-paths perspective (though honestly the code path is not very complicated on the mailroom side either way). But honestly we would probably want to continue to include `contact_id` and `group_id` relations on FlowStart just so that we can logically represent things to the user. (or maybe not and we say we are educating them by showing only a query instead of the contact id and group they entered)

Option 2 also doubly makes us depend on our indexes being not too stale and actually correct. This PR is already making that more the case and honestly Elastic seems to have treated us pretty well in that respect, but adding it as a dependency for every single start feels kinda scary.

Thoughts? @rowanseymour @ericnewcomer @norkans7 ?

hooks/session_triggered.go
nicpottier: Ya, I started that way too but then that first list of required args is kinda long and seemed more readable this way to me. 🤷‍♂

rowanseymour: IMHO looks a bit weird to split a package and symbol. What if `DoRestartParticipants` was also `WithRestartParticipants(bool)` and then you also don't have to add new bool types?

nicpottier: Went back and updated those.

ericnewcomer:
I don't think I hate a query being the canonical form for flow starts. In a way it might actually be clearer to the user when we ultimately surface flow starts. ie, that the selection of contacts was a point-in-time that matched a query instead of it just being a group list. I mean, it is the same thing, but in a way I can see that feeling like a more honest visualization of what happened.

So, in the whatsapp case, is the thinking they would create a query that is `in this group AND not in this other group`? As an aside, kind of like the idea of surfacing in group queries, but having them be uuids isn't particularly exciting.

nicpottier:
Ya, we'd have to come up what our syntax is like for group queries and that could be using a group name:
  `age=10 AND group = "U-Reporters"`
Or
  `age = 10 AND group != "U-Reporters"`

Seems mostly clear..

But ya, the case above for whatsapp would then be something like:

`group = "U-Reporters" AND whatsapp = ""`

ericnewcomer:
Then of course they change their group name, doh. Maybe that's okay for that link to be tenuous in the flow start case. Assuming we wouldn't allow in-group querying for general contact search (or at least disable dynamic group creation from those queries).

nicpottier:
Ya, maybe we rewrite to UUID and store that version in our own queries?

rowanseymour:
👿 devils advocate: you don't need to query on groups.. every group should be dynamic anyway.. backed by a query on fields.. everything is fields

nicpottier:
I don't necessarily hate that, believe Vumi did that. It doesn't feel super friendly for users though, so feel it may require us hiding this from them in the simplest case. But ya, then suddenly you get to do boolean operations on all your groups which is powerful.

models/assets.go
```diff
@@ -417,3 +422,54 @@ func (a *OrgAssets) ResthookBySlug(slug string) *Resthook {
...
+		if policy == envs.RedactionPolicyURNs {
+			field.Category = search.Unavailable
+		} else {
+			field.Category = search.Scheme
+		}
+		return field
+	}
+
+	// is this an implicit field?
+	if key == contactql.ImplicitKey {
```
rowanseymour: We don't actually allow implicit conditions (`Bob` instead of `name = Bob`) in dynamic group queries - you can search like that but you can't save that raw query as a group. Thus goflow's evaluation of groups errors if there's an implicit condition. Reason for that is that the RP UI should rewrite any implicit conditions anyway - so searching for `Bob` gets rewritten as the explicit condition `name ~ "Bob"`. Seems like it would be reasonable to do same with these kinds of queries and then you don't have to worry about implementing implicit queries.

search/search.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	var query elastic.Query
+
+	if field.Category == Implicit {
+		number, _ := strconv.Atoi(c.Value())
+
+		if field.Key == NameTel {
+			if number != 0 {
+				return elastic.NewNestedQuery("urns", elastic.NewMatchPhraseQuery("urns.path", number)), nil
```
rowanseymour: see comment above re implicit conditions, but doesn't this need to be restricted to tel urns?

models/assets.go
nicpottier: Sure, that's true for dynamic queries but didn't see any reason why this part of the library shouldn't support it.

search/search.go
nicpottier: Oops, yes indeed, will update.

models/assets.go
rowanseymour: Why wouldn't we rewrite these queries same as we do for contact search?

rowanseymour: Also to be clear, it's not just implicit conditions we rewrite away, it's also implicit ANDs, e.g. `Bob 078245353` becomes `name ~ "Bob" AND tel ~ 078245353`

nicpottier: Oh I think we will totally pass in preprocessed ones. But this module only concerns itself with taking a contactql query and turning that into something elastic supports. Implicit queries are a thing in contactql, so seemed correct to support them.

Yes it is a superset of what we will absolutely need, but it is implemented and tested so don't see the plus in ripping that out.

rowanseymour: It's a superset of what will _actually be run_. Your call but I don't think we've tended to merge extra complexity just because it's already written and tested.

nicpottier: You don't have to love it! :)

rowanseymour: Yeah as I said your call. I just can't imagine you making the same argument on someone else's PR.

nicpottier: It is a module, we have all sorts of utility modules that support more than we strictly run. To me it is clearest to have it support everything contactql does. I'm not doing this to spite you, promise you.

rowanseymour: I don't think you are - just that we generally don't commit code unless there's a use for it.. but in this case I guess this might be used in the future if we delegate all contact searching to mailroom, so all good. Would be great to have one implementation of contactql in one language.

search/search.go
```diff
@@ -0,0 +1,365 @@
...
+func conditionToElasticQuery(env envs.Environment, registry FieldRegistry, c *contactql.Condition) (elastic.Query, error) {
+	field := registry.LookupSearchField(c.Key())
+	if field == nil {
+		return nil, errors.Errorf("unable to find field: %s", c.Key())
+	}
+
+	var query elastic.Query
+
+	if field.Category == Implicit {
+		number, _ := strconv.Atoi(c.Value())
```
rowanseymour: should this match the logic of https://github.com/nyaruka/rapidpro/blob/5814c169e031656c1f65e97348e382d90c496c54/temba/contacts/search/parser.py#L1053 ?

i.e. match a phonenumber regex, then do a bit of normalization?

models/assets.go
rowanseymour: Tho actually looking at the Python contactql parser, it always switches out an implicit condition with an explicit id, name or tel ones _in the visitor_, so if we want to be consistent, goflow's visitor should also do that, and then there is no `contactql.ImplicitKey`.. this code can't happen.

The only reason goflow's implementation doesn't already do that is we've been assuming it would never encounter an implicit condition.

nicpottier: I guess the question is what is ContactQL? Does ContactQL support implicit queries? If so, then I would posit everything that takes contactql should support implicit queries. If not, then that should be an error when parsing contactql and it is the job of RapidPro to turn implicit looking things into valid ContactQL.

I'm fine with either, though to me it seems useful to have implicit queries be part of ContactQL.

rowanseymour: Implicit conditions are certainly a part of contactql - but our Python implementation swaps them out with explicit ones at the visitor level so the evaluator has no concept of an implicit condition. No reason for goflow to not do likwise.

Also noticed that goflow's parser is wrong for implicit keys - it sets the operator to `=` instead of `~`.. it's never been used 🤷‍♂ 

Have made https://github.com/nyaruka/goflow/pull/742 which I think replicates the behavior of the Python parser

nicpottier: Excellent, so then every place that takes ContactQL can take any ContactQL, which I think is the right thing.

rowanseymour: Ok just made goflow v0.44.3 which should in theory match the python parser!

search/search.go
```diff
@@ -0,0 +1,267 @@
...
+// Field represents a field that elastic can search against
+type Field interface {
+	assets.Field
+	UUID() uuids.UUID
+}
+
+// FieldRegistry provides an interface for looking up queryable fields
+type FieldRegistry interface {
+	LookupSearchField(key string) Field
+}
```
nicpottier, line 24: So this and the interface above are only needed because `assets.Field` doesn't have a UUID parameter. (and we don't have a `FieldUUID` in goflow either)

How do you feel about adding that? Would make this all fall away to just one field resolver func.

models/fields.go
```diff
@@ -15,7 +15,7 @@ import (
 type FieldID int
 
 // FieldUUID is our type for the UUID of a field
-type FieldUUID uuids.UUID
+type FieldUUID = uuids.UUID
```
nicpottier: Bit of a hack to get around the circular dependency between `models` and `search`. If we had a `FieldUUID` in assets then wouldn't need this.

(spent a few hours trying to refactor all ids and uuids in `models` to an `ids` package but it just got out of control, when do we get decent golang refactor tools again, fuck modules)

search/search.go
rowanseymour: Took me a moment to remember what field UUID is used for.. but yeah we could add it to assets.Field

search/search.go
```diff
@@ -0,0 +1,267 @@
...
+	key := c.PropertyKey()
+
+	if c.PropertyType() == contactql.PropertyTypeField {
+		field := registry.LookupSearchField(key)
+		if field == nil {
+			return nil, errors.Errorf("unable to find field: %s", key)
+		}
+
+		fieldQuery := elastic.NewTermQuery("fields.field", field.UUID())
+		fieldType := field.Type()
```
rowanseymour, line 65: so no need to look up anything to do with fields if we add UUID right? FieldType is also on `contactql.Condition`

nicpottier: Right, ya, then we'd have everything we need to build these queries in `assets.Field`

search/search.go
rowanseymour: https://github.com/nyaruka/goflow/pull/750

nicpottier:
@rowanseymour updated, oh god please no more changes. :)

search/search.go
```diff
@@ -0,0 +1,262 @@
...
+
+	return elastic.NewBoolQuery().Should(queries...), nil
+}
+
+func conditionToElasticQuery(env envs.Environment, resolver contactql.FieldResolverFunc, c *contactql.Condition) (elastic.Query, error) {
+	var query elastic.Query
+	key := c.PropertyKey()
+
+	if c.PropertyType() == contactql.PropertyTypeField {
+		field := resolver(key)
```
rowanseymour, line 59: is this needed? haven't we already verified that the fields exist during parsing

rowanseymour: oh sorry I get it now - there's still no field UUID on the condition

rowanseymour (approved):
Can't say I'm familiar with the ES API to know if that's all right, but rest looks good.

Any way to verify if this works the same as the ES stuff in RP ?

nicpottier:
I've run these queries against Elastic and it is pretty much a translation of the Rapid queries so I think they should be good, won't know for sure until we try!

PR #146, opened 2019-07-11
deal with brand new URNs when sending messages

### LINKED ISSUE #145: Adding URN and changing Channel in same flow fails
TL;DR-
Condition- Create a flow that both adds a new URN and also tries to change that new channel to the preferred one.
Results- new URN for the new channel is added, but the old channel remains the default, no error logged
Expected- new channel should begin being used immediately after changing the priority in the flow.

Longer:

A flow that:
1) Updates Contact with a new URN (telegram)
2) sends a message on the old channel (SMS) confirming
3) Sets the channel to telegram
4) sends a new message (telegram)

In this case, both messages (2 and 4) go over SMS, and SMS remains the default channel.

From what I can tell in the mailroom code, the adding of the urn (contact_urn_changed) is batch processed at the end of the session so when step 3 happens as a channel event, it tries to update the urn priority but it silently fails because the new URN hasn't been added yet because the session is still ongoing.
two things:
 shouldn't there be a specific channel event type for changechannel (which would still call UpdatePreferredURN) but would log an error instead of ignoring a noop? Since we're specifically asking for a channel priority change, the inability to change it is a failure
shouldn't any major contact changes (like adding a URN) be made synchronously in the flow so it's available to other ops?

nicpottier:
Fixes: https://github.com/nyaruka/mailroom/issues/145

### DIFF
--- a/hooks/msg_created.go
+++ b/hooks/msg_created.go
@@ -190,7 +190,7 @@ func handleMsgCreated(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, org *mod
 		return nil
 	}
 
-	// messaging flows must have urn id set on them, if not, go look it up
+	// messages in messaging flows must have urn id set on them, if not, go look it up
 	if session.SessionType() == models.MessagingFlow {
 		urn := event.Msg.URN()
 		if models.GetURNInt(urn, "id") == 0 {
--- a/hooks/msg_created.go
+++ b/hooks/msg_created.go
@@ -190,11 +190,16 @@ func handleMsgCreated(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, org *mod
 		return nil
 	}
 
-	// messaging flows must have urn id set on them, assert that
+	// messaging flows must have urn id set on them, if not, go look it up
 	if session.SessionType() == models.MessagingFlow {
-		urnInt := models.GetURNInt(event.Msg.URN(), "id")
-		if urnInt == 0 {
-			return errors.Errorf("attempt to create messaging message with 0 id URN")
+		urn := event.Msg.URN()
+		if models.GetURNInt(urn, "id") == 0 {
+			urn, err := models.GetOrCreateURN(ctx, tx, org, session.ContactID(), event.Msg.URN())
+			if err != nil {
+				return errors.Wrapf(err, "unable to get or create URN: %s", event.Msg.URN())
+			}
+			// update our Msg with our full URN
+			event.Msg.SetURN(urn)
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


models/contacts.go
```diff
@@ -975,14 +993,14 @@ WHERE
...
-	Display   *string   `db:"display"`
-	Auth      *string   `db:"auth"`
-	Scheme    string    `db:"scheme"`
-	Priority  int       `db:"priority"`
-	OrgID     OrgID     `db:"org_id"`
+	ContactID ContactID   `db:"contact_id"`
+	Identity  string      `db:"identity"`
+	Path      string      `db:"path"`
+	Display   null.String `db:"display"`
+	Auth      null.String `db:"auth"`
```
nicpottier, line 1000: cleanup, pointer strings are the devil

models/events.go
```diff
@@ -33,7 +33,7 @@ func ApplyPreEventHooks(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, org *O
 	for hook, args := range preHooks {
 		err := hook.Apply(ctx, tx, rp, org, args)
 		if err != nil {
-			return errors.Wrapf(err, "error applying pre commit hook: %v", hook)
+			return errors.Wrapf(err, "error applying pre commit hook: %T", hook)
```
nicpottier, line 36: @rowanseymour not sure if you knew about `%T` but it cracks interfaces and prints the underlying type name which is mighty useful.

models/msgs.go
```diff
@@ -295,25 +291,22 @@ func NewOutgoingMsg(orgID OrgID, channel *Channel, contactID ContactID, out *flo
...
 		m.ChannelUUID = channel.UUID()
 		msg.SetChannelID(channel.ID())
-	}
-	m.MsgCount = 1
-
-	// get the id of our URN
-	urnInt := GetURNInt(out.URN(), "id")
-	if urnInt != 0 {
-		urnID := URNID(urnInt)
-		m.ContactURNID = &urnID
```
nicpottier, line 313: This is done in `SetURN`

hooks/msg_created.go
```diff
@@ -190,11 +190,16 @@ func handleMsgCreated(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, org *mod
 		return nil
 	}
 
-	// messaging flows must have urn id set on them, assert that
+	// messaging flows must have urn id set on them, if not, go look it up
```
rowanseymour: comment sounds like the flow needs an id

rowanseymour (approved):

PR #148, opened 2019-07-15
🔥  fix leaking db conns

nicpottier:
We are leaking db conns (almost certainly via a missing `rows.Close()`) causing a deadlock after a while. Found a few possible spots after looking at all our uses of `Query*`, URN related ones are most likely culprits since that recently changed (though they aren't new). Other possibility is the auth endpoint, which may be triggered by different usage all of a sudden (more surveyor submissions), the check of auth tokens was leaking a conn on every hit.

### DIFF
--- a/models/channel_connection.go
+++ b/models/channel_connection.go
@@ -141,6 +141,7 @@ func InsertIVRConnection(ctx context.Context, db *sqlx.DB, orgID OrgID, channelI
 	if err != nil {
 		return nil, errors.Wrapf(err, "error inserting new channel connection")
 	}
+	defer rows.Close()
 
 	rows.Next()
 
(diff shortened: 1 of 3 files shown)


norkans7:
I suspect something around IVR retries too

nicpottier:
One of the fixes was related to IVR actually. Why do you suspect that?

norkans7:
error report pointed to L285 which seems to have the defer already

norkans7 (approved):

nicpottier:
Ya, these are tricky to track down because errors you will see are actually where a healthy thread is trying to grab a connection. I did a thread dump on a locked up mailroom to verify some of these. The auth token was the one I could see in the trace.

PR #152, opened 2019-07-23
Update to goflow v0.43.0

rowanseymour:


handler/handler_test.go
```diff
@@ -6,14 +6,15 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gomodule/redigo/redis"
```
rowanseymour, line 9: `goimports` orders imports in 3 sections (std lib, your own, third party) so I've always copied that.. any objections?

nicpottier: You can do whatever you like as long as you don't make me do that kind of shit work. :)

nicpottier (approved):

PR #155, opened 2019-07-26
🚀 Latest goflow

rowanseymour:


handler/worker.go
```diff
@@ -400,13 +400,18 @@ func HandleChannelEvent(ctx context.Context, db *sqlx.DB, rp *redis.Pool, eventT
...
-	var params types.XValue
+	var params *types.XObject
 	if event.Extra() != nil {
 		asJSON, err := json.Marshal(event.Extra())
 		if err != nil {
 			return nil, errors.Wrapf(err, "unable to marshal extra from channel event")
 		}
-		params = types.JSONToXValue(asJSON)
+		if len(asJSON) > 0 {
+			params, err = types.ReadXObject(asJSON)
```
nicpottier: What if `ReadXObject` returned nil when passed a len(0) `json.RawSummary`? Would remove the outer if block from all these.

nicpottier (approved):

handler/worker.go
rowanseymour: Oddity here is that `event.Extra()` is a map which we marshal to JSON and then load again with `types.ReadXObject`. So if it was `nil` then `json.Marshal(event.Extra())` would return `"null"`.

nicpottier: Wouldn't `len()` still return 0 though? Maybe I'm misunderstanding. I'm not suggesting always marshalling, but having a special switch at the very top of `ReadXObject` that check for `len(x) == 0` and returns nil in that case.

rowanseymour: Maybe I'm misunderstanding. What I mean is that if you remove the outer if here, then extra might be `nil` that get's serialized as `"null"`, so `ReadXObject` would blow up.

But we could just remove the _inner_ `if` here because if extra is non-nil, then it's impossible for `len(asJSON) > 0` to be false.

rowanseymour:
@nicpottier updated to write session UUIDs

handler/worker.go
nicpottier: If `types.ReadXObject` started with:
```golang
func types.ReadXObject(data json.RawMessage) *XObject {
  if len(data) == 0 { return nil }
...
```

Then couldn't you make the above (and everywhere else):
```golang
params, err := types.ReadXObject(start.Extra())
if err != nil {
  return errors.Wrap(err, "unable to read JSON from flow start extra")
}
```

rowanseymour: `event.Extra()` isn't `json.RawMessage` tho.. it's `map[string]interface{}`

nicpottier: Err, sorry, I guess I added this to the wrong one.. the others are `json.RawMessage` though no?

nicpottier (approved):

handler/worker.go
rowanseymour: So I guess that feels like I might be breaking some go convention. Seems like there's an unwritten rule that for a function that returns `(X, error)`, either `X` is non-nil or `error` is non-nil.. but never neither and never both.. no? 

I know it means an extra `if` but it seems more go-ey to say `ReadX` returns an `X` or an `error`. I dunno. Have you seen any examples in the std lib of that not being the case?

nicpottier: Oh that's an interesting point. I would say the convention is more that the value is invalid if `err` is returned, not that `nil` values have any specific meaning.

I think there are probably lots of cases where methods read `nil` but no error. Any kind of search/lookup is naturally going to lead to that kind of interface. `nil` isn't particularly special in go (remember it is still a function receiver!) so I don't think there's any rule there really.

nicpottier: But I approved it! Just don't love it!

nicpottier: I would say my bias here kinda comes from seeing Fitz and others like using `nil` receivers for example to simplify outer code. IE, they seem to have a bias to cleverness in struct implementations that simplifies outer concerns. And this just felt like a case of that.

rowanseymour: As much as I want you to love this.. I think I'm gonna hit that merge button on the basis that we already have a `ReadX` method for everything in goflow and their contracts are always "return an X or an error". So if nothing else, this is consistent.

nicpottier: It is an X, just a nil X. :) Merge away though I hate the repetition you are adding to mailroom with your ideals.

nicpottier: Could get rid of an `if` block there too:
  https://github.com/nyaruka/goflow/blob/ce4ec8e51a8e4b8e1844dad1a0c28bb01ef52619/flows/triggers/base.go#L161

:)

PR #165, opened 2019-08-14
Handle FlowRun having nil Flow, use FlowReference instead

nicpottier:


go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/mattn/go-sqlite3 v1.10.0 // indirect
 	github.com/nyaruka/ezconf v0.2.1
 	github.com/nyaruka/gocommon v1.1.1
-	github.com/nyaruka/goflow v0.47.0
```
rowanseymour, line 27: good job we're still < 1.0 and the go module police can't get us for all these breaking changes!

rowanseymour (approved):

PR #166, opened 2019-08-15
✏️ Start writing flows_flowrun.status alongside exit_type

rowanseymour:


models/runs.go
```diff
@@ -33,19 +33,36 @@ type FlowRunID int64
 const NilFlowRunID = FlowRunID(0)
 
 const (
-	SessionStatusActive      = "A"
```
rowanseymour, line 36: removing this because mailroom shouldn't ever be writing a session with this status - "active" is used in internally within goflow but any session being handed back to a caller must be in a "waiting" or one of the ended states.

rowanseymour: at least I think that's right.. @nicpottier ?

models/runs.go
```diff
@@ -210,6 +233,7 @@ type FlowRun struct {
 	r struct {
 		ID         FlowRunID     `db:"id"`
 		UUID       flows.RunUUID `db:"uuid"`
+		Status     RunStatus     `db:"status"`
```
nicpottier, line 236: I *think* this is fine because Mailroom doesn't read FlowRuns, only writes them, but just considering that this will blow up if we do ever read a flow run with a `null` status.

nicpottier (approved):

PR #169, opened 2019-08-21
⌚️ Add base schedule model, add next fire functionality and test

nicpottier:
Functionality to move schedule fires to mailroom.

This will need to be merged / shipped at same time as https://github.com/nyaruka/rapidpro/pull/2532


models/triggers.go
```diff
@@ -285,7 +289,8 @@ FROM
 WHERE 
 	t.org_id = $1 AND 
 	t.is_active = TRUE AND
-	t.is_archived = FALSE
+	t.is_archived = FALSE AND
+	t.trigger_type != 'S'
```
nicpottier, line 293: Wasn't causing any issues I don't think but from mailroom's perspective there aren't the kind of triggers it wants to load and evaluate on orgs when receiving messages etc..

tasks/schedules/cron.go
```diff
@@ -0,0 +1,123 @@
...
+		tz, err := s.Timezone()
+		if err != nil {
+			log.WithError(err).Error("error firing schedule, unknown timezone")
+			continue
+		}
+
+		// if it is a broadcast
+		if s.Broadcast() != nil {
+			// clone our broadcast, our schedule broadcast is just a template
+			bcast := models.CloneBroadcast(s.Broadcast())
```
nicpottier: Note that we do not insert a brand new broadcast here. This matches what we do in flows when triggering broadcasts in that we only create DB objects for broadcasts for user generated broadcasts, but this IS different than what we currently do in RapidPro. (which similarly used to create broadcast objects for flows)

tasks/schedules/cron.go
```diff
@@ -0,0 +1,123 @@
...
+				log.WithError(err).Error("error firing scheduled broadcast")
+				continue
+			}
+			broadcasts++
+
+		} else if s.FlowStart() != nil {
+			start := s.FlowStart()
+
+			// insert our flow start
+			err := models.InsertFlowStarts(ctx, db, []*models.FlowStart{start})
```
nicpottier: Conversely we DO need co insert flow starts because IVR starts require a DB objects to look up attributes in the start when getting IVR start callbacks. This also matches flow engine behavior in mailroom.

nicpottier:
Looks like we do expose fired broadcasts in the UI so I'm going to add inserting those with the parent.

Screen Shot 2019-09-23 at 11 47 11 AM

go.mod
```diff
@@ -41,3 +41,5 @@ require (
 	gopkg.in/go-playground/validator.v9 v9.21.0
 	gopkg.in/mail.v2 v2.3.1 // indirect
 )
+
+go 1.13
```
rowanseymour: is this cuz you're running 1.13 locally?

models/starts.go
```diff
@@ -196,8 +196,13 @@ type startGroup struct {
 // InsertFlowStarts inserts all the passed in starts
 func InsertFlowStarts(ctx context.Context, db Queryer, starts []*FlowStart) error {
 	is := make([]interface{}, len(starts))
-	for i := range starts {
-		is[i] = &starts[i].s
+	for i, s := range starts {
+		// populate UUID if needbe
```
rowanseymour: sp

rowanseymour (approved):

nicpottier:
@rowanseymour can you take another look? Now inserting cloned broadcasts for that case.

models/msgs.go
```diff
@@ -487,14 +489,16 @@ type BroadcastTranslation struct {
 // Broadcast represents a broadcast that needs to be sent
 type Broadcast struct {
 	b struct {
-		BroadcastID   BroadcastID                             `json:"broadcast_id,omitempty"`
+		BroadcastID   BroadcastID                             `json:"broadcast_id,omitempty" db:"id"`
```
rowanseymour, line 492: this does make for a slightly weird struct since some fields map to the database representation.. some don't, e.g. `template_state` is stored in `msgs_broadcast.metadata`

models/msgs.go
```diff
@@ -517,16 +522,136 @@ func NewBroadcast(
...
+func InsertChildBroadcast(ctx context.Context, db Queryer, parent *Broadcast) (*Broadcast, error) {
+	child := NewBroadcast(
+		parent.OrgID(),
+		NilBroadcastID,
+		parent.b.Translations,
+		parent.b.TemplateState,
+		parent.b.BaseLanguage,
+		parent.b.URNs,
+		parent.b.ContactIDs,
+		parent.b.GroupIDs,
```
rowanseymour, line 545: I guess we currently don't need to clone things like attachments or quick replies, but might be something we enable in the future.

rowanseymour (approved):
lets hope this code lasts 5 years!

models/msgs.go
nicpottier: Oh ya for sure, mailroom's models are definitely optimized for particular paths which is why most methods are very specific in what they do. Without a real ORM its kinda the only practical way to go but ya you have to be careful in how you use said models. (good example is the insanity of the expired schedules query which is essentially the equiv of `.select_related()` but all custom SQL)

models/msgs.go
nicpottier: I'll add an assertion for those not being set.

PR #170, opened 2019-08-26
📱 Use DTOne for airtime transfers if configured

rowanseymour:
Although this might now be called DTone or Thune.. still not sure and still don't have credentials for testing


nicpottier:
I think the abstraction is right but not sure if this belongs in mainroom as opposed to either goflow or maybe something else. This and other functionality like this should be reused, we don't want anybody who implements airtime transfers to have to reimplement it from scratch every time, same for any other provider.

So I think most of this probably belongs in goflow which then leans on mailroom to instantiate the provider appropriately based on org level configs etc..

rowanseymour:
The problem with putting it in goflow is if we want/need to make database changes. For example if we want to replicate existing transferto functionality in the legacy engine then we should be saving a `airtime_airtimetransfer` row for each transfer. 

Is there really much reusable code here anyway? Not sure how you'd reuse any of this if you were working with a completely different airtime provider.

nicpottier:
Is saving an `airtime_airtimetransfer` any different than us saving a webhook log?

It isn't that the code would be used with another airtime provider, it's that another thing may want to wrap goflow like mailroom does and use transferto. This will be especially true for NLP providers.

rowanseymour:
@nicpottier oh yeah that does make sense re saving those when handling the event. Ok let me have a go at abstracting the config so mailroom can do that resolving but goflow owns the parts doing the work

models/env.go
```diff
@@ -6,13 +6,27 @@ import (
...
 )
 
+const (
+	configDTOneLogin    = "TRANSFERTO_ACCOUNT_LOGIN"
+	configDTOneToken    = "TRANSFERTO_AIRTIME_API_TOKEN"
+	configDTOnecurrency = "TRANSFERTO_ACCOUNT_CURRENCY"
+)
+
+func init() {
+	goflow.SetServiceResolver(&serviceResolver{})
```
rowanseymour, line 27: bit weird but `models` imports `goflow` so `goflow` can't import `models`

Issue #171, opened 2019-08-27
Update interrupt task to support interrupting runs by flow


rowanseymour:
RP needs to trigger this when a flow is archived

PR #172, opened 2019-08-28
⛈️ Allow interrupting sessions by flow

rowanseymour:
Addresses #171 and also took it upon myself to make the test a bit more robust


interrupts/worker.go
```diff
@@ -45,6 +46,16 @@ WHERE
...
 `
 
+const activeSessionIDsForFlowsSQL = `
+SELECT 
+	id
+FROM 
+	flows_flowsession fs
+WHERE
+	fs.status = 'W' AND
+	fs.current_flow_id = ANY($1);
```
rowanseymour, line 56: @nicpottier aware you might not see this for a while but I think this is correct... interrupt only sessions where we're _currently_ in the flow. If we have a parent run from the flow being interrupted we'll error at runtime if we try to resume back into it (it'll have been archived).

PR #179, opened 2019-09-25
add support for triggering sessions via query within a flow

nicpottier:


hooks/session_triggered.go
```diff
@@ -69,8 +69,8 @@ func (h *InsertStartHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool
 			event := e.(*events.SessionTriggeredEvent)
 
 			// we skip over any session starts that involve groups if we are in a batch start
-			if len(sessions) > 1 && len(event.Groups) > 0 {
-				logrus.WithField("session_id", s.ID).Error("ignoring session trigger on group in batch")
+			if len(sessions) > 1 && (len(event.Groups) > 0 || event.ContactQuery != "") {
```
nicpottier, line 72: Note that footgun prevention is extending to starting by query. I think that's right but open to discussion.

rowanseymour: just making sure I'm understanding this correctly - `len(sessions) > 1` would only be the case when we're starting a batch of contacts in a flow, and processing the events from the first sprint - as a batch - so this would be an attempt to start a group immediately before the first wait?

nicpottier: Correct, this is a batch start of some kind (either manually or from a flow) that then before any wait is triggering another session.

rowanseymour: Ok makes sense treating queries same as groups here

go.mod
```diff
@@ -41,3 +41,5 @@ require (
 	gopkg.in/go-playground/validator.v9 v9.21.0
 	gopkg.in/mail.v2 v2.3.1 // indirect
 )
+
+go 1.13
```
rowanseymour, line 45: what is this?

rowanseymour (approved):

go.mod
nicpottier: Just an annotation of the version of go. Looks like technically enables interpretation of new literal formats: https://golang.org/doc/go1.13#language

rowanseymour: yeah sorry was more wondering what having this in go.mod actually means.. like if goflow's go.mod has it.. can it only be used in a parent project that also has it? what does it mean for testing against 1.12. .etc

PR #181, opened 2019-10-10
Add PostgreSQL 11 tests

brianhlin:
PG11 wouldn't start without the `postgres-11-postgis-2.5-scripts` package and this other hack: https://stackoverflow.com/questions/55161807/travis-ci-not-connecting-to-postgresql-11-2.

It looks like the tests themselves are failing with PG11, though. Unfortunately, I'm not familiar enough with Go or the test suite to understand if it's actually a problem or just an issue with the tests :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ go:
 matrix:
   include:
     - name: "Postgres 11"
+      env:
+        - PGDB=11
       addons:
         postgresql: "11"
         apt:
@@ -16,6 +18,8 @@ matrix:
             - postgresql-client-11
             - postgresql-11-postgis-2.5-scripts
     - name: "Postgres 10"
+      env:
+        - PGDB=10
       addons:
         postgresql: "10"
         apt:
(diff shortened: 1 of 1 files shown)


nicpottier:
Looked into this some and it looks like the problem is that pg11 can't read dumps from pg10 and our `mailroom_test.dump`

It *does* look like dumps from 11 work fine in 10 though. So we'll just have to create those there. I'll update this PR with that dump in a sec.

rowanseymour:
@nicpottier I'm updating the dump in https://github.com/nyaruka/mailroom/pull/183 with PG10

nicpottier:
Gonna have to be pg11 then to pass both.

nicpottier:
Bugger, I was being trigged by the binaries on my dev box. Reverting all this. We're going to test with 10 only on mailroom.

nicpottier:
Finally figured out the problem here. Dump needs to be created in 11 and that'll work across 10 and 11. So we are testing both now:
  https://github.com/nyaruka/mailroom/pull/186

Added assertion that the dump is being created in right version here:
  https://github.com/nyaruka/rapidpro/pull/2569

PR #182, opened 2019-10-10
😼  NLU support

nicpottier:
* [x] latest goflow
* [x] update mailroom_db with test classifiers
* [x] reading / testing classifiers
* [x] hook for classifier events
* [x] classifier service factory
* [x] writing of http logs
* [x] http log tests
* [x] hook tests


go.sum
```diff
@@ -73,6 +73,10 @@ github.com/nyaruka/gocommon v1.1.1 h1:RnQ+kMzN1lA+W0NpkBDd0mGU3UqadJygR3SMpITMYT
 github.com/nyaruka/gocommon v1.1.1/go.mod h1:QbdU2J9WBsqBmeZRuwndf2f6O7rD7mkC0bGn5UNnwjI=
 github.com/nyaruka/goflow v0.50.4 h1:71sGdTg6Ia7T/1uPPMUVo20clyW5rA8/f0JRXjTDRP0=
 github.com/nyaruka/goflow v0.50.4/go.mod h1:tPppKrlURChn6oRUpCVTzM39f6kH3exSSsFjTXzIGTw=
+github.com/nyaruka/goflow v0.51.0 h1:F4aVgeo5oA2N7iC+bwtJp9vx8RMaW2zKeBIOWhfxeuw=
+github.com/nyaruka/goflow v0.51.0/go.mod h1:wI+5qZRajxxZvfYXqbmWauTWAJ0GOTTG3TZDkixU/E0=
+github.com/nyaruka/goflow v0.52.0 h1:QYiXBdKq8zwYQTDhsZ5YZyuHcvZ8S4ar9tUuU2kdJbE=
+github.com/nyaruka/goflow v0.52.0/go.mod h1:wI+5qZRajxxZvfYXqbmWauTWAJ0GOTTG3TZDkixU/E0=
```
rowanseymour, line 74: go mod tidy

goflow/engine.go
```diff
@@ -31,7 +38,10 @@ func Engine() flows.Engine {
 	engInit.Do(func() {
 		eng = engine.NewBuilder().
 			WithHTTPClient(httpClient).
-			WithWebhookService(webhooks.NewService("RapidProMailroom/"+config.Mailroom.Version, 10000)).
+			WithWebhookServiceFactory(func(flows.Session) (flows.WebhookService, error) {
```
rowanseymour: there is a `webhooks.NewServiceFactory` if you want it

hooks/classifier_called.go
```diff
@@ -0,0 +1,82 @@
...
+		"status":          event.Status,
+		"elapsed_ms":      event.ElapsedMS,
+		"classifier_name": event.Classifier.Name,
+		"classifier_uuid": event.Classifier.UUID,
+	}).Debug("classifier called")
+
+	// if this is a connection error, use that as our response
+	response := event.Response
+	if event.Status == flows.CallStatusConnectionError {
+		response = "connection error"
```
rowanseymour: This is a little weird.. since then we have some response traces which are valid HTTP traces and some that aren't. Could we leave this blank or encode this state in some other way? Or maybe `is_error=True AND response=None` is enough? If we're including this for display reasons then arguably it should be something localizable. 

Anyways.. don't feel _that_ strongly if you want to keep it this way

rowanseymour (approved):

hooks/classifier_called.go
nicpottier: Ya, it is a little odd for sure. I don't hate the `is_error` and null response solution. I'll do that.

go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/mattn/go-sqlite3 v1.10.0 // indirect
 	github.com/nyaruka/ezconf v0.2.1
 	github.com/nyaruka/gocommon v1.1.1
-	github.com/nyaruka/goflow v0.50.4
+	github.com/nyaruka/goflow v0.52.0
```
rowanseymour: pfft so yesterday.. v0.52.1 is where it's at now

PR #183, opened 2019-10-15
💸 Use DTOne for airtime transfers if configured

rowanseymour:
Replaces #170


hooks/airtime_transferred.go
```diff
@@ -0,0 +1,115 @@
...
+type InsertAirtimeTransfersHook struct{}
+
+var insertAirtimeTransfersHook = &InsertAirtimeTransfersHook{}
+
+// Apply inserts all the airtime transfers that were created
+func (h *InsertAirtimeTransfersHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, org *models.OrgAssets, sessions map[*models.Session][]interface{}) error {
+	// gather all our transfers and logs
+	transfers := make([]*models.AirtimeTransfer, 0, len(sessions))
+	logs := make([]*models.HTTPLog, 0, len(sessions))
+	logsByTransfer := make(map[*models.AirtimeTransfer][]*models.HTTPLog, len(sessions))
```
rowanseymour: this seem right where I need to insert somethings, then insert other things which reference them?

hooks/broadcast_created_test.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func TestBroadcastCreated(t *testing.T) {
-	testsuite.ResetRP()
+	testsuite.Reset()
```
rowanseymour, line 19: this used to be the first hook test

hooks/airtime_transferred.go
nicpottier: I think the steps are right ya. Seems like maybe you could remove the map though?

IE:
1. build up the single list of airtime transfers
2. insert them
3. iterate across the list of transfers, building a single list of logs (setting the airtime transfer id in the process)
4. insert them

rowanseymour: True will update!

nicpottier (approved):

models/airtime.go
```diff
@@ -0,0 +1,110 @@
...
+type AirtimeTransfer struct {
+	t struct {
+		ID            AirtimeTransferID     `db:"id"`
+		OrgID         OrgID                 `db:"org_id"`
+		Status        AirtimeTransferStatus `db:"status"`
+		ContactID     ContactID             `db:"contact_id"`
+		Sender        null.String           `db:"sender"`
+		Recipient     urns.URN              `db:"recipient"`
+		Currency      null.String           `db:"currency"`
+		DesiredAmount null.String           `db:"desired_amount"`
```
rowanseymour: NB treating decimals as strings here.. which works because we're only writing them and don't need them back as decimals from the DB

hooks/airtime_transferred_test.go
```diff
@@ -0,0 +1,127 @@
...
+			httpx.NewMockResponse(200, "error_code=13\r\nerror_txt=Oops\r\n"),
+		},
+	}))
+
+	testsuite.DB().MustExec(
+		`UPDATE orgs_org SET config = '{"TRANSFERTO_ACCOUNT_LOGIN": "nyaruka", "TRANSFERTO_AIRTIME_API_TOKEN": "123456789"}'::jsonb
+		WHERE id = $1`, models.Org1)
+
+	tcs := []HookTestCase{
+		HookTestCase{
```
rowanseymour, line 86: big fan of this test setup

nicpottier: the SQL assertions are a bit awkward / laborious but ya, I like the flow running side of it.

nicpottier: FYI, I often use two different contacts in the same "TestCase" doing different things. This doubly tests that the rollup inserts are doing what you expect.

rowanseymour: So curiously enough if I change the contact ID to george on the second hooktestcase.. it seems to save the transfer as cathy... I'm sorta stumped and I can dig into this more.. but is there anything obvious I'm doing wrong here..?

nicpottier: Hrmm.. that could be a bug in the hooks test. Not sure I've ever done one with only George (usually it's Cathy + other people). The way that test works is it tries to split by the contact ID right at the start and only that contact gets the actions but maybe that test/operand is busted.

PR #185, opened 2019-10-16
🚀 Update to goflow v0.53.0

rowanseymour:


hooks/classifier_called.go
```diff
@@ -43,34 +43,37 @@ func (h *InsertHTTPLogsHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.P
...
+			httpLog.URL,
+			httpLog.Request,
+			httpLog.Response,
+			httpLog.Status != flows.CallStatusSuccess,
+			httpLog.ElapsedMS,
+			httpLog.CreatedOn,
+		)
 
-	session.AddPreCommitEvent(insertHTTPLogsHook, log)
+		session.AddPreCommitEvent(insertHTTPLogsHook, log)
```
rowanseymour, line 76: is it better do this... or change the event type to a list of logs? Assuming it doesn't make much difference in this case

models/http_logs.go
```diff
@@ -50,7 +48,7 @@ func NewClassifierCalledLog(
 	h.h.Request = request
 	h.h.Response = null.String(response)
 	h.h.IsError = isError
-	h.h.RequestTime = int(elapsed / time.Millisecond)
+	h.h.RequestTime = elapsedMS
```
rowanseymour: the event has time in millis.. so feels weird to convert to a duration just to create this.. then convert back to millis

hooks/classifier_called.go
nicpottier: They all get wrapped up into a list in the end anyways, so probably clearest to do single items.

models/http_logs.go
nicpottier: This I disagree with. Durations should be passed as durations in go interfaces. Can understand if we have  JSON objects that want to convert to MS, but interfaces should be typed.

nicpottier (changes requested):

models/http_logs.go
rowanseymour: The problem is that the value isn't a golang duration (count of nanoseconds).. it's duration as a count of milliseconds. It's seems weird to convert milliseconds to nano-seconds and back to milliseconds just so you can have a type in this function call.. and obscures the fact we're throwing away accuracy. 

I'll change it back if you really want but if it's the typing your after then I think we just have something like `type Millis int`

nicpottier: Its a duration, the resolution of that duration (based on how it was measured or in this case how it was saved to an event) doesn't really apply. The interface to create a log should take a duration, regardless of the resolution of that duration. That it gets converted to milliseconds when written to the DB is an implementation detail that should be hidden through the interface.

This one I really do feel strongly about, so yes, please change to `time.Duration`

rowanseymour: Changed

nicpottier: Gracias.

PR #191, opened 2019-10-22
🤑 Simulation shouldn't really send airtime

rowanseymour:
Simulation should use a different engine instance with a fake airtime service

### DIFF
--- a/goflow/engine.go
+++ b/goflow/engine.go
@@ -6,16 +6,33 @@ import (
 	"sync"
 	"time"
 
+	"github.com/nyaruka/gocommon/urns"
 	"github.com/nyaruka/goflow/flows"
 	"github.com/nyaruka/goflow/flows/engine"
 	"github.com/nyaruka/goflow/services/webhooks"
 	"github.com/nyaruka/mailroom/config"
+
+	"github.com/shopspring/decimal"
 )
 
-var eng flows.Engine
-var engInit sync.Once
+var httpClient *http.Client
+var eng, simulator flows.Engine
+var engInit, simulatorInit sync.Once
 var classificationFactory engine.ClassificationServiceFactory
 var airtimeFactory engine.AirtimeServiceFactory
+
+func init() {
+	// customize the default golang transport
+	t := http.DefaultTransport.(*http.Transport).Clone()
+	t.MaxIdleConns = 32
+	t.MaxIdleConnsPerHost = 8
+	t.IdleConnTimeout = 30 * time.Second
+	t.TLSClientConfig = &tls.Config{
+		Renegotiation: tls.RenegotiateOnceAsClient, // support single TLS renegotiation
+	}
+
+	httpClient = &http.Client{Transport: t, Timeout: time.Duration(15 * time.Second)}
+}
 
 // RegisterClassificationServiceFactory can be used by outside callers to register a classification factory
 // for use by the engine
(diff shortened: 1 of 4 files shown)


goflow/engine.go
```diff
@@ -57,3 +60,43 @@ func Engine() flows.Engine {
...
+func (s *simulatorAirtimeService) Transfer(session flows.Session, sender urns.URN, recipient urns.URN, amounts map[string]decimal.Decimal, logHTTP flows.HTTPLogCallback) (*flows.AirtimeTransfer, error) {
+	transfer := &flows.AirtimeTransfer{
+		Sender:        sender,
+		Recipient:     recipient,
+		DesiredAmount: decimal.Zero,
+		ActualAmount:  decimal.Zero,
+	}
+
+	// pick arbitrary currency/amount pair in map
+	for currency, amount := range amounts {
```
rowanseymour, line 94: the alternative to this would be giving the DTOne service a param that tells it to do everything except the actual transfer. That would mean..

-  You need DTOne configured on your org or you see an error in the simulation
- We correctly resolve the currency and actual amount that _would_ be sent

I'm on the fence...

nicpottier: Seems to dovetail with us gating these features based on resources being added, which seems to be the direction we are going. If we never show the action when there's no provider displayed then we could indeed make the calls to figure out what we would send etc..

That said, I'm totally fine with this and it gives us the possibility of choosing otherwise. (IE, letting people build flows without a provider connected)

nicpottier (approved):
I like it! Plus makes me doubly like the service model.
(this will also be nice for the email case btw)

goflow/engine.go
rowanseymour: Ok I'm gonna merge this so we can push DTOne stuff (i.e. just need to make sure simulation doesn't make real transfers)... but I'm leaning toward the second option

rowanseymour:
@nicpottier yeah services working out nicely thus far.. also see https://github.com/nyaruka/mailroom/issues/192

nicpottier:
I think the ideal in the end is basically both. IE, if we there is a provider configured we use that to get what the transfer would be, otherwise we just fake it. 

I *do* think there is potential for us deciding from a product perspective it is useful to tease these features, that would just require us to be a bit more explicit in the editor as to what you need to do to add them or whatnot. So ideally we wouldn't put ourselves in a box where they couldn't try these features out without the required dependencies. (gets trickier for classifiers though)

rowanseymour:
If you take transfers as an example tho and say we let users add transfer actions without having a provider configured.. there's not that much difference between seeing a simulator message that says _no airtime service configured_ and _we fake sent $1_... and maybe the former is actually more useful 🤷‍♂

nicpottier:
Sorry, not sure I understand the distinction. What's the former you are referring to?

rowanseymour:
Former is if we used DTOne service for simulation.. initialized in such a way that it knows not to really to the transfer. You get real currency/amount resolving.. but it requires the user to have DTOne configured.

Second is this PR.

Just saying that from a flow building POV.. you don't don't gain a whole lot by having simulated transfers "succeed".. maybe its more useful to surface an explanation to the user in the editor that they need to go add a provider on their account page.

nicpottier:
Oh I see. Ya, honestly I'm not sure what side I fall on. We also have to consider the import case and what it means then..

PR #195, opened 2019-11-01
🧪 Use github actions

rowanseymour:


.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Install Redis
+      uses: zhulik/redis-action@v1.0.0
+      with:
+        redis version: '5'
+
+    - name: Install PostgreSQL
+      uses: nyaruka/postgis-action@master
+      with:
+        postgresql version: '${{ matrix.pg-version }}'
+        postgresql init scripts: testdb
```
rowanseymour: you can pass in db name/username/password here but then the whole database is owned by that user, which doesn't work when later mailroom does `DROP OWNED BY mailroom_test CASCADE`

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+      with:
+        redis version: '5'
+
+    - name: Install PostgreSQL
+      uses: nyaruka/postgis-action@master
+      with:
+        postgresql version: '${{ matrix.pg-version }}'
+
+    - name: Install Linux packages
+      run: sudo apt install -y --no-install-recommends postgresql-client
```
rowanseymour: so ideally you wouldn't install a client because that's one more thing that can break when versions change, and the docker image used by our `nyaruka/postgis-action` action above in theory supports having a directory of init scripts used to initialize the db... but I haven't as yet been able to get that to work.. or see why it's not working

.github/workflows/test.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    - name: Install Redis
+      uses: zhulik/redis-action@v1.0.0
+      with:
+        redis version: '5'
+
+    - name: Install PostgreSQL
+      uses: nyaruka/postgis-action@master
+      with:
+        postgresql version: '${{ matrix.pg-version }}'
```
rowanseymour: Can pass username/database/password here.. but if you pass `mailroom_test` etc, then that becomes the superuser and the tests fail because they rely on `DROP OWNED BY mailroom_test` to reset the database between tests.

testsuite/testsuite.go
```diff
@@ -44,7 +44,7 @@ func ResetDB() {
 		dir = path.Dir(dir)
 	}
 
-	mustExec("pg_restore", "-d", "mailroom_test", "-U", "mailroom_test", path.Join(dir, "./mailroom_test.dump"))
+	mustExec("pg_restore", "-h", "localhost", "-d", "mailroom_test", "-U", "mailroom_test", path.Join(dir, "./mailroom_test.dump"))
```
rowanseymour, line 47: need host so it doesn't try to use a local socket file

nicpottier:
What's the story on releases?

nicpottier:
It'll be really interesting to see if these end up being more reliable. At first blush they have a bit of node feel to them in that there are so many outside dependencies on actions that could disappear at a moment's notice. But looks good! Just need to figure out our goreleaser story before switching full bore.

rowanseymour:
The thing about actions is that you can just copy them into `.github/actions` or just host them yourself like https://github.com/nyaruka/postgis-action

I had ripped out the release stuff whilst testing but will try putting it back

nicpottier:
Ya, understood, though if I'm not mistaken that postgis-action has an external dependency as well! :)

rowanseymour:
@nicpottier can you tell from looking at https://github.com/nyaruka/mailroom/runs/289759286 if it *would've* worked with a real release number? Not sure I can prove this works without making a real release.

nicpottier (approved):

nicpottier:
Oh for sure, let's give it a go!

PR #197, opened 2019-11-06
👩🏽‍🚀 Update to latest goflow

rowanseymour:


models/flows.go
```diff
@@ -242,15 +202,14 @@ SELECT ROW_TO_JSON(r) FROM (SELECT
 	coalesce(metadata, '{}')::jsonb as config,
 	definition::jsonb || 
 		jsonb_build_object(
-			'name', f.name,
 			'uuid', f.uuid,
-			'flow_type', f.flow_type,
```
rowanseymour: think I need to tweak this a bit more - had to remove `flow_type` because if that's in the definition then goflow starts treating the definition as a legacy definition...

- need to decide what the criteria is for goflow deciding a definition is legacy... is it the presence of certain keys like `flow_type` or the absence of keys like `spec_version`
- ideally this SQL wouldn't be creating a merge of the new and legacy specs cuz that gets confusing later.. i.e. only add `metadata` if it sees the version is < 12

@nicpottier ?

runner/runner.go
```diff
@@ -740,7 +740,7 @@ func TriggerIVRFlow(ctx context.Context, db *sqlx.DB, rp *redis.Pool, orgID mode
 func validateFlow(sa flows.SessionAssets, uuid assets.FlowUUID) error {
 	flow, err := sa.Flows().Get(uuid)
 	if err != nil {
-		return errors.Wrapf(err, "invalid flow: %s, cannot start", flow.UUID())
```
rowanseymour, line 743: this was a panic if err != nil

web/simulation/simulation.go
```diff
@@ -267,10 +267,7 @@ func populateFlow(org *models.OrgAssets, uuid assets.FlowUUID, flowDef json.RawM
 	}
 
 	if legacyFlowDef != nil {
-		err = flow.SetLegacyDefinition(legacyFlowDef)
```
rowanseymour, line 270: right now flows can be passed in as `legacy_definition` or `definition` makes no difference

models/flows.go
nicpottier: Looking at something like `spec_version` feels most correct to me. If not there then legacy, if it is then that version?

web/simulation/simulation.go
nicpottier: Should probably change signature then to take just one def ya?

rowanseymour: well this is a way to keep things working until editor is updated

nicpottier: Oh I see, because you don't want to do that switch above. ok.

nicpottier (approved):

Issue #200, opened 2019-11-19
Engine timeouts for HTTP calls 


rowanseymour:
Currently the engine uses one HTTP client for all services (webhooks, NLP, airtime) and its timeout is set to 15 seconds.

https://github.com/nyaruka/mailroom/blob/8db69f332f282a1366e97b92e7b3da99bf32a6c8/goflow/engine.go#L25-L34

DT One transfers (at least to Ecuador) are timing out after 15 seconds and their support guy suggested I need to increase that to 600 seconds (their own maximum)

Pretty sure we don't want to use 600 for all HTTP connections from the engine.. not sure we even want to do that for DT One.. but seems like services should have their own HTTP clients and thus different timeouts.

nicpottier:
10 MINUTE HTTP REQUESTS!?! :)

rowanseymour:
Lol yeah.. I mean no... but we could bump the DT One to a minute. It could be just the carrier here is really slow? I need to find more pre-payers to test with.

nicpottier:
Anyways, agree that we probably need separate clients for these. I'm pretty sure our fair queuing should take care of a large blast doing lots of these not totally melting us.

rowanseymour:
Added in #201.. DT One gets 2 minutes.. everyone still gets 15 seconds

Issue #192, opened 2019-10-22
Use header to indicate webhook calls are made during simulation


rowanseymour:
As of #191 simulation uses a different goflow engine instance which provides a simple way to have simulated sessions use different services.. and webhooks are a service. So instead of the goflow webhook service taking a User-Agent string, it would take a map of default headers to add to any request.

Which could include `X-RapidPro-Simulation` header.. giving users with their own webhook services the option to handle simulator calls differently.

nicpottier:
Ya this is a nice thing for sure.

rowanseymour:
Added in #201

PR #201, opened 2019-11-20
💿 Latest goflow 

rowanseymour:
To address #200 and #192


models/env.go
```diff
@@ -18,9 +19,12 @@ import (
 
 // Register a airtime service factory with the engine
 func init() {
+	// give airtime transfers an extra long timeout
+	httpClient := &http.Client{Timeout: time.Duration(120 * time.Second)}
```
rowanseymour, line 23: how does 2 minutes sound for DT One?

goflow/engine.go
```diff
@@ -64,9 +68,13 @@ func Engine() flows.Engine {
 // Simulator returns the global engine instance for use with simulated sessions
 func Simulator() flows.Engine {
 	simulatorInit.Do(func() {
+		webhookHeaders := map[string]string{
+			"User-Agent":      "RapidProMailroom/" + config.Mailroom.Version,
+			"X-Mailroom-Mode": "simulation",
```
rowanseymour, line 73: So many different ways to convey that the request is part of simulation.. could be part of the User-Agent.. should it be called Mailroom or Temba or RapidPro.. opted for this over `X-Mailroom-Simulation: 1` as it's more extendable.

models/assets.go
```diff
@@ -439,3 +439,7 @@ func (a *OrgAssets) ResthookBySlug(slug string) *Resthook {
 func (a *OrgAssets) Templates() ([]assets.Template, error) {
 	return a.templates, nil
 }
+
+func (a *OrgAssets) Globals() ([]assets.Global, error) {
+	return nil, nil // not supported yet
```
rowanseymour, line 444: need to add the models on the django side, then wire this up, then add the UI on the django side

rowanseymour:
@nicpottier note the coverage change for `hooks/email_created`.. which I assume is unrelated but seems that test isn't doing what it used to do

rowanseymour:
@nicpottier actually that's back to normal.. an intermittent thing evidently

models/env.go
nicpottier: Better than 10 minutes? :)

nicpottier (approved):

models/env.go
rowanseymour: I just can't believe an HTTP request ever came back after 10 minutes.. I can't

PR #202, opened 2019-11-20
@ Globals

rowanseymour:


models/globals.go
```diff
@@ -0,0 +1,71 @@
...
+	g.key as key,
+	g.name as name, 
+	g.value as value
+FROM 
+	globals_global g
+WHERE 
+	org_id = $1 
+ORDER BY 
+	key ASC
+) r;
```
nicpottier, line 71: Should this have `is_active = TRUE` since this is a smart model?

nicpottier (approved):

models/globals.go
rowanseymour: good question.. so on the RP side it's currently hard deleting these so is_active is always true.. but I am filtering on is_active=True any time I read them.. so is it best to just consistently ignore that field if we're not using it? Or check it for all models..

nicpottier: If you are filtering then probably best to be consistent even if never used. In any case can't imagine any scenario where we would WANT mailroom to show inactive ones.

PR #208, opened 2019-12-17
⚙️ Flow migrations

rowanseymour:
Update to latest goflow to add support for flow migrations


web/flow/testdata/validate_valid.response.json
```diff
@@ -1,7 +1,7 @@
 {
     "uuid": "8f107d42-7416-4cf2-9a51-9490361ad517",
     "revision": 106,
-    "spec_version": "13.0.0",
+    "spec_version": "13.1.0",
```
rowanseymour: this endpoint is gonna be removed https://github.com/nyaruka/rapidpro/pull/2639

web/flow/testdata/migrate_minimal_legacy.json
```diff
@@ -31,6 +31,5 @@
             "name": "No ruleset flow",
             "saved_on": "2015-11-20T11:02:19.790131Z"
         }
-    },
-    "include_ui": true
```
rowanseymour, line 35: we removed this a while ago - UI is always included now

go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/nyaruka/mailroom
 
 require (
+	github.com/Masterminds/semver v1.4.2
```
ericnewcomer, line 4: Masterminds!

nicpottier:
Looks good but do you really want to merge this just before the holidays?

rowanseymour:
I feel like tomorrow gives me enough time to test this (and also I'm not going anywhere next week). The problem is I already merged the goflow side of it thinking this would be a quick merge.. and I don't like having things left in a state where we can't make hotfixes if need be over the holidays.

PR #212, opened 2020-01-08
Webhook retries

rowanseymour:


goflow/engine.go
```diff
@@ -94,6 +88,27 @@ func Simulator() flows.Engine {
...
+		t.TLSClientConfig = &tls.Config{
+			Renegotiation: tls.RenegotiateOnceAsClient, // support single TLS renegotiation
+		}
+
+		webhooksHTTPClient = &http.Client{
+			Transport: t,
+			Timeout:   time.Duration(config.Mailroom.WebhooksTimeout) * time.Second,
+		}
+
+		webhooksHTTPRetries = httpx.NewRetryDelays(3, 10)
```
rowanseymour: how to make this configurable 🤔

nicpottier: Maybe only make the delay (pre-jitter) configurable? 3 tries ought to be enough for anybody...

rowanseymour: Ok how about..

```
WebhooksTimeout: 15000
WebhooksInitialBackoff: 5000  / / doubled for each WebhooksRetries retry
WebhooksRetries: 2
```

which would mean _try for 15 secs, wait 5 secs, try for 15 secs, wait 10 secs, try for 15 secs_

config/config.go
```diff
@@ -50,17 +55,22 @@ type Config struct {
...
+		Elastic:        "http://localhost:9200",
+		BatchWorkers:   4,
+		HandlerWorkers: 32,
+		LogLevel:       "error",
+		Version:        "Dev",
+
+		WebhooksTimeout:        15000,
+		WebhooksMaxRetries:     2,
+		WebhooksInitialBackoff: 5000,
+		WebhooksBackoffJitter:  0.5,
```
rowanseymour, line 70: so with these settings you get 3 attempts in total separated by backoffs of [3.75,6.25] and [7.5,12.5] seconds.. probably enough?

nicpottier: Given that we've never really done this at all before sure seems like it!

nicpottier: I don't care much but I don't feel this deserves a config value.

nicpottier (approved):

PR #213, opened 2020-01-09
Mailroom contact search API

nicpottier:
Adds new endpoint to search for contacts within an org given a root group and the query. Returns the page of contact ids required, total number of hits and parsed query.


README.md
```diff
@@ -71,7 +71,7 @@ To run the tests you need to create the test database:
 
 ```
 $ createdb mailroom_test
-$ createuser -P -E temba (set no password)
+$ createuser -P -E -s mailroom_test (set no password)
```
nicpottier, line 74: new machines have you discover things!

models/contacts_test.go
```diff
@@ -46,11 +46,19 @@ func TestElasticContacts(t *testing.T) {
 				"query":{
 					"bool":{
 						"must":[
-							{"match":{"name":{"query":"george"}}},
-							{"term":{"org_id":1}},
-							{"term":{"is_active":true}},
-							{"term":{"is_blocked":false}},
-							{"term":{"is_stopped":false}}
+							{ "bool":{
```
nicpottier, line 49: Queries are a bit different now since the building of the root query doesn't force `blocked` and `stopped` anymore. (since those are valid searches)

search/search.go
```diff
@@ -254,9 +293,22 @@ func conditionToElasticQuery(env envs.Environment, resolver contactql.FieldResol
...
+			return nil, NewError("unsupported scheme comparator: %s", c.Comparator())
 		}
 	}
 
-	return nil, errors.Errorf("unsupported property type: %s", c.PropertyType())
+	return nil, NewError("unsupported property type: %s", c.PropertyType())
+}
+
+// Error is used when an error is in the parsing of a field or query format
+type Error struct {
```
nicpottier, line 304: `search.Error` is the type for search errors that should be shown to the user and are therefore (400 status errors). All other types of errors will result in the contact search endpoint returning a 500 (IE, elastic is unreachable)

search/search_test.go
```diff
@@ -23,6 +24,54 @@ func (f *MockField) Name() string           { return f.fieldKey }
...
+		{"descending date", "-dob", `{"fields.datetime":{"nested":{"filter":{"term":{"fields.field":"cbd3fc0e-9b74-4207-a8c7-248082bb4572"}},"path":"fields"},"order":"desc"}}`, nil},
+		{"descending state", "-state", `{"fields.state":{"nested":{"filter":{"term":{"fields.field":"67663ad1-3abc-42dd-a162-09df2dea66ec"}},"path":"fields"},"order":"desc"}}`, nil},
+		{"ascending district", "district", `{"fields.district":{"nested":{"filter":{"term":{"fields.field":"54c72635-d747-4e45-883c-099d57dd998e"}},"path":"fields"},"order":"asc"}}`, nil},
+		{"ascending ward", "ward", `{"fields.ward":{"nested":{"filter":{"term":{"fields.field":"fde8f740-c337-421b-8abb-83b954897c80"}},"path":"fields"},"order":"asc"}}`, nil},
+
+		{"unknown field", "foo", "", fmt.Errorf("unable to find field with name: foo")},
+	}
+
+	resolver := func(key string) assets.Field {
+		return registry[key]
```
rowanseymour: obviously doesn't matter in this case but my eyes have been trained to spot stuff like this (if item does't exist, nil struct gets wrapped in non-nil interface)

rowanseymour (approved):

search/search_test.go
nicpottier: Ah, ya, I'm not as trained to look for those yet, thanks for the reminder.

PR #215, opened 2020-01-10
Add config option to max bytes of a webhook call response body

koallann:
When configure a webhook call in a flow, there is a validation for size of response, if the size body is greater than 10k bytes, the mailroom can't get the responses.

For example: using bothub, some responses is greater than 10k.

image

This pr makes the value customizable via environment variable. When raise the mailroom service, can set a value for accept and treat wehbooks call.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -24,6 +24,7 @@ type Config struct {
 
 	WebhooksTimeout        int     `help:"the timeout in milliseconds for webhook calls from engine"`
 	WebhooksMaxRetries     int     `help:"the number of times to retry a failed webhook call"`
+	WebhooksMaxBodyBytes   int     `help:"the maximum size of bytes to a webhook call response body"`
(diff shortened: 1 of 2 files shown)


rowanseymour:
This is probably fine.. but just upping that value isn't a great solution as those larger webhook bodies end up in the session JSON, negatively effecting parsing times etc. We've been discussing at https://github.com/nyaruka/goflow/issues/814 ways of keeping larger payloads around but not in the session JSON.

matmsa27:
Cool @rowanseymour , we opened this PR to a solution for now, we agree with your explanation and is the best way to improve this point.

Can we contribute to this solution with you? We are available to help.

nicpottier:
I CC'ed @johncordeiro on a few issues that we would accept contributions for. The big change to webhook handling is a tricky one so that's one we would handle ourselves.

rowanseymour (approved):

PR #219, opened 2020-01-22
Add endpoint to parse contactql query

nicpottier:
Needed for validating queries for starts as well as for dynamic group editing.

Thought about getting cute and making this a `GET` operation on search (or maybe even `HEAD`) but that seems less clear.

### DIFF
--- a/web/contact/contact_test.go
+++ b/web/contact/contact_test.go
@@ -20,7 +20,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestServer(t *testing.T) {
+func TestSearch(t *testing.T) {
 	testsuite.Reset()
 	ctx := testsuite.CTX()
 	db := testsuite.DB()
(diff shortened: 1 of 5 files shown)


web/contact/contact_test.go
```diff
@@ -154,3 +164,72 @@ func TestServer(t *testing.T) {
...
+		Status int
+		Error  string
+		Query  string
+		Fields []string
+	}{
+		{
+			"/mr/contact/parse_query", "GET",
+			"",
+			405, "illegal method: GET",
+			"", nil,
```
rowanseymour, line 194: As a general practice, I think it's useful to write tests comparing the actual JSON returned so you can see stuff like if an array field is sometimes `null` instead of `[]` or omitted.. (in this case I just wanna be sure `"fields"` isn't `null` sometimes.

rowanseymour (approved):

web/contact/contact_test.go
nicpottier: As in a literal string of the JSON? I'm parsing the response as JSON and then comparing the actual fields here which feels like the right level of detail to me and less fragile to ordering changes etc..

rowanseymour: You can use `assert.JSONEq` or `test.AssertJSONEqual` (I wrote the latter to get better diffs).. those ignore ordering changes. The problem with parsing out specific fields is you might not catch something being `null` vs `[]`, or new fields appearing.

nicpottier: Makes sense, not going to rewrite these but will look at using that on the next JSON endpoint I do.

PR #221, opened 2020-01-24
allow dynamic group population within mailroom

nicpottier:
- [x] deal with campaign event deletion creation.. arr


nicpottier:
Random note that we probably don't want to allow dynamic groups based on group queries or we will have crazy cascading results.

rowanseymour (approved):
Still trying to wrap my head around all around all the possible race conditions but they're not new

PR #222, opened 2020-01-28
Update to latest goflow

rowanseymour:


rowanseymour:
@nicpottier I've found one more little thing to tweak in goflow - disallowing `!=` for number and date fields and catching that at parse time too.. but just wanna check.. we really can't support that in ES?

nicpottier:
Can't I don't know for sure, but don't currently. Would have to do some research to figure all that out and whether there is a reason we don't support it now.

rowanseymour:
Merging this for now as it's a lot more correct than it was. Also note there's a drop in coverage because some evaluation errors are now technically impossible because they're caught at the parser stage.

PR #239, opened 2020-02-14
Convert flow and expression endpoint testing to be snapshot based

rowanseymour:
Every time we change the output of `/flow/inspect` just have to run the tests with `-write`


nicpottier:
Was thinking the other day how most of our web test cases could all be the same. Any reason to not just have these in a single file that specifies all the attributes? (IE, json list of dicts that contain URL, method, body, expected status, expected body?)

rowanseymour:
@nicpottier that sounds great - tho might not work for everything because some of the tests do things like check counts in the database.. tho guess even those can be described as JSON.. some do other things like pass part of the response from the last call (simulator).. I'd like to merge this for now as a step in that direction if that's ok

rowanseymour:
And codecov has just reminded me that probably still best to keep a separate tests file for each package otherwise you don't get the coverage

nicpottier:
With respect to checking DB etc, that's all the same as this ya? I don't see any special casing there. But ya, sure can merge.

nicpottier (approved):

rowanseymour:
Not sure I follow.. these are just request/response checks. I didn't convert tests like https://github.com/nyaruka/mailroom/blob/master/web/surveyor/surveyor_test.go#L55 cuz they include SQL assertions

nicpottier:
Yep yep, sorry I had interpreted you bringing that stuff up as a downside to the full file approach as opposed to this one, my mistake.

PR #240, opened 2020-02-14
Asset cleanup

nicpottier:
Bit of refactoring around how our assets are loaded etc..

Callers can now request a cached org asset but can specify which pieces of the cache to refresh. This comes in handy for those endpoints that absolutely must have the most current stuff.

Also need some more thought around mucking with flows since new contact action endpoint needs to run "fake" flows in production. Decided to add a `Clone` method that builds new assets but doesn't cache the result (and bases it off an existing env). Some runtime panics make sure you don't try to modify uncloned assets.


models/assets.go
```diff
@@ -363,26 +473,21 @@ func (a *OrgAssets) FlowByID(flowID FlowID) (*Flow, error) {
...
 
 // SetFlow sets the flow definition for the passed in ID. Should only be used for unit tests
-func (a *OrgAssets) SetFlow(flowID FlowID, flow flows.Flow) (*Flow, error) {
-	// build our definition
-	definition, err := json.Marshal(flow)
-	if err != nil {
-		return nil, errors.Wrapf(err, "error marshalling flow definition")
+func (a *OrgAssets) SetFlow(id FlowID, uuid assets.FlowUUID, name string, definition json.RawMessage) *Flow {
+	if !a.cloned {
+		panic("can only override flow definitions on cloned orgs")
```
nicpottier, line 477: I don't necessarily love this approach.. the better one would be to have a different type of org that satisfied the org interface but had this way of being modified, but that would have exceeded my budget for refactoring for the day and would require a lot of copy/paste.

models/flows.go
```diff
@@ -68,11 +68,6 @@ func (f *Flow) FlowType() FlowType { return f.f.FlowType }
 // Version returns the version this flow was authored in
 func (f *Flow) Version() string { return f.f.Version }
 
-// SetDefinition sets our definition from the passed in definition
```
nicpottier, line 71: We want all our assets to be immutable. This is now handled by calling `SetFlow` on OrgAssets.

ivr/ivr.go
```diff
@@ -339,14 +339,8 @@ func StartIVRFlow(
...
 
-	// build our session assets
-	sa, err := models.GetSessionAssets(org)
-	if err != nil {
-		return errors.Wrapf(err, "error starting flow, unable to load assets")
-	}
-
 	// our flow contact
-	contact, err := c.FlowContact(org, sa)
+	contact, err := c.FlowContact(org)
```
nicpottier, line 343: A lot of signature changes like this as session assets are now cached on the org which makes the calls a lot simpler in many places. SessionAssets are thread safe and immutable ya @rowanseymour ?

rowanseymour: Yeah the only cleverness is trying to cache flow defs

nicpottier: Is the caching thread safe?

rowanseymour: In theory https://github.com/nyaruka/goflow/blob/master/flows/definition/assets.go#L31

rowanseymour (approved):

PR #242, opened 2020-02-14
Modify Contact Endpoint

nicpottier:
Adds an endpoint that accepts a subset of actions to modify a contact.

Returns the full modified contact and the list of events applied.


web/contact/contact.go
```diff
@@ -177,3 +187,196 @@ func handleParseQuery(ctx context.Context, s *web.Server, r *http.Request) (inte
...
+
+// Request update a contact. Clients should only pass in the fields they want updated.
+//
+//   {
+//     "org_id": 1,
+//     "contact_uuid": "559d4cf7-8ed3-43db-9bbb-2be85345f87e",
+//     "name": "Joe",
+//     "fields": {
+//        "age": "124"
+//     },
```
rowanseymour: needs updated

web/contact/contact.go
```diff
@@ -177,3 +187,196 @@ func handleParseQuery(ctx context.Context, s *web.Server, r *http.Request) (inte
...
+	actions.TypeAddContactGroups: true,
+	actions.TypeAddContactURN:    true,
+	// actions.TypeRemoveContactURN  <-- missing
+	actions.TypeRemoveContactGroups: true,
+	actions.TypeSetContactChannel:   true,
+	actions.TypeSetContactLanguage:  true,
+	actions.TypeSetContactName:      true,
+	actions.TypeSetContactTimezone:  true,
+	actions.TypeSetContactField:     true,
+}
```
nicpottier: So.. the slight bummer is we already have the concept of things that /modify/ contacts, which we created for surveyor. In that case it is the set of modifications that took place on the contact which we then apply after we write the session. (we only want to apply the diffs to a contact when a surveyor submission comes in)

Those would be nice to use, but the problem is that there's a good bit of logic in the actions themselves that is desirable to keep, namely how we parse fields and URNs etc.. 

@rowanseymour any thoughts there? I'd be fine rejigging this to use modifiers instead but we'd have to address those things.

web/contact/contact.go
```diff
@@ -177,3 +187,196 @@ func handleParseQuery(ctx context.Context, s *web.Server, r *http.Request) (inte
...
+type applyActionsResponse struct {
+	Contact *flows.Contact `json:"contact"`
+	Events  []flows.Event  `json:"events"`
+}
+
+// the types of actions our apply_actions endpoind supports
+var supportedTypes map[string]bool = map[string]bool{
+	actions.TypeAddContactGroups: true,
+	actions.TypeAddContactURN:    true,
+	// actions.TypeRemoveContactURN  <-- missing
```
rowanseymour: https://github.com/nyaruka/goflow/issues/860

web/contact/testdata/apply_actions.json
```diff
@@ -0,0 +1,207 @@
...
+        },
+        "status": 200,
+        "response": {
+            "contact": {
+                "uuid": "6393abc0-283d-4c9b-a1b3-641a035c34bf",
+                "id": 10000,
+                "timezone": "America/Los_Angeles",
+                "created_on": "2020-02-10T10:57:50.710449-08:00",
+                "urns": [
+                    "tel:0788555111"
```
nicpottier: I'm a bit confused here, shouldn't "add urn" normalize the URN? Don't see that done in the action.. seems like a bug.

web/contact/contact_test.go
```diff
@@ -319,3 +323,82 @@ func TestParse(t *testing.T) {
...
+	dates.SetNowSource(dates.NewSequentialNowSource(time.Date(2018, 7, 6, 12, 30, 0, 123456789, time.UTC)))
+	defer dates.SetNowSource(dates.DefaultNowSource)
+
+	// for simpler tests we clear out cathy's fields and groups to start
+	db.MustExec(`UPDATE contacts_contact SET fields = NULL WHERE id = $1`, models.CathyID)
+	db.MustExec(`DELETE FROM contacts_contactgroup_contacts WHERE contact_id = $1`, models.CathyID)
+	db.MustExec(`UPDATE contacts_contacturn SET contact_id = NULL WHERE contact_id = $1`, models.CathyID)
+
+	type TestCase struct {
+		Label        string          `json:"label"`
```
nicpottier: Just playing around with the single testcase to rule them all here, will pull out once I feel it makes sense.

nicpottier:
This is pretty much there save for more tests.. kinda pausing until we discuss whether we want to use modifiers instead.

web/contact/contact.go
rowanseymour: It definitely seems like.. conceptually modifiers are the thing to use.. As they exist currently:

 1. They run outside of the context of a run or session
 2. They against data which has already been evaluated/resolved/parsed - i.e. all expressions are gone, names of groups resolved to real groups etc
 3. They run against data which is assumed to be valid/trusted

It's kind of funny to think about point 2  if we went with actions.. and users started typing expressions into the contact editing UI.

Point 3 does mean there is a burden on RP to only send valid data.. Maybe that's reasonable because the RP UI should be a source of trustworthy data.. maybe we can push some validation down into modifiers...? Can brainstorm in a call if you want.

nicpottier: I feel validation would probably be nice to have, only because one of the points here is trying to reduce the number of places we do that kind of thing. So having validation / parsing done in only one place seems desirable. Then again, maybe this stuff isn't that complicated so maybe it is ok?

Cases I can think of where there is real parsing being done that has any kind of logic:
 1) dates, since their format is org specific
 2) locations
 3) URNs
 
Not the end of the world to duplicate those, just a bit of a shame. If we were starting from scratch I think I'd pick modifiers doing that parsing / error handling and actions being in charge of expression evaluation.

rowanseymour: Modifiers were just a thing we invented to solve a different problem so it's ok to stop and re-evaluate what their role should be going forward. If we feel they are a better fit than actions (even if they're not a perfect fit).. I think it's ok to say that there's a little bit of duplication of validation for now and we look at tweaking modifiers to do more of that work in future, and be a better fit for both of these usecases.

I'm almost curious to see you try to implement this with modifiers just so we can get a clearer grasp on how much duplication we're talking about. Also worth remembering all of this logic is currently duplicated in Python.. so _less_ duplication would still be a win.

nicpottier: Well to be clear I'm not going to parse the modifiers then try to validate those on the mailroom side, the validation would be on the python side as it builds the modifiers.

The mailroom implementation will basically look like it does now, just with modifiers instead. And I need to move onto other things to tracking down all the Python places where we modify contacts and porting it to using this endpoint will have to be a team effort.

models/events.go
```diff
@@ -9,68 +9,79 @@ import (
 	"github.com/pkg/errors"
 )
 
-// EventCommitHook defines a callback that will accept a certain type of events across session, either before or after committing
-type EventCommitHook interface {
-	Apply(context.Context, *sqlx.Tx, *redis.Pool, *OrgAssets, map[*Session][]interface{}) error
+// Scene represents the context that events are occurring in
+type Scene struct {
```
nicpottier, line 13: So did a bunch of refactoring so we could use all the hooks we have without the concept of a session. Problem is that they do need a concept of the context of the events they are in and well, we have used a lot of those words up. (context, session, environment)

Enter Scene! Events happen in a scene, which could be built from a session or from a contact. No I don't love love it, but I like it better than anything else I could come up with.

nicpottier:
Now modifiers @rowanseymour

web/contact/contact_test.go
```diff
@@ -319,3 +323,85 @@ func TestParse(t *testing.T) {
 		test.AssertEqualJSON(t, tc.Response, json.RawMessage(response), "%d: unexpected response", i)
 	}
 }
+
+var update = flag.Bool("update", false, "update testdata files")
```
rowanseymour: all the goflow snapshot tests use `-write` but I prefer `-update` so I'll change them for consistency

hooks/airtime_transferred.go
```diff
@@ -25,11 +25,11 @@ type InsertAirtimeTransfersHook struct{}
 var insertAirtimeTransfersHook = &InsertAirtimeTransfersHook{}
 
 // Apply inserts all the airtime transfers that were created
-func (h *InsertAirtimeTransfersHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, org *models.OrgAssets, sessions map[*models.Session][]interface{}) error {
+func (h *InsertAirtimeTransfersHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, org *models.OrgAssets, scenes map[*models.Scene][]interface{}) error {
```
rowanseymour, line 28: this hook is using `scenes` plural, others are using `scene` singular for the same thing.. plural seems right

web/contact/contact_test.go
nicpottier: Found `update` as the switch in a bunch of articles talking about:
  https://medium.com/@jarifibrahim/golden-files-why-you-should-use-them-47087ec994bf

That said, I don't name my files `.golden` so obviously I'm not totally bought in.

hooks/airtime_transferred.go
nicpottier: Doh, rename / refactor turds. Fixed.

PR #243, opened 2020-02-21
Add support for searching by group

rowanseymour:
Pausing to go learn ES..


web/contact/contact.go
```diff
@@ -125,12 +125,14 @@ type parseRequest struct {
...
 //   "fields": ["age"],
-//   "elastic_query": { .. }
+//   "elastic_query": { .. },
+//   "allow_as_group": true
 // }
 type parseResponse struct {
 	Query        string      `json:"query"`
 	Fields       []string    `json:"fields"`
 	ElasticQuery interface{} `json:"elastic_query"`
+	AllowAsGroup bool        `json:"allow_as_group"`
```
rowanseymour, line 135: seems like we shouldn't trust RP on this.. and maybe even push this decision down to goflow

web/contact/testdata/parse_query.json
```diff
@@ -28,11 +28,11 @@
         "path": "/mr/contact/parse_query",
         "body": {
             "org_id": 1,
-            "query": "age \u003e 10"
+            "query": "AGE>10"
```
rowanseymour, line 31: make sure we're sending back a rewritten query

search/search.go
```diff
@@ -14,6 +14,9 @@ import (
 	"github.com/shopspring/decimal"
 )
 
+// GroupResolverFunc resolves a group name to a UUID because ES indexes groups by the latter
+type GroupResolverFunc func(name string) assets.Group
```
nicpottier: What about having a single interface that has a `ResolveField` and `ResolveGroup` function on it? Just seems we are building lots of anonymous functions to represent something instead of just having an interface that would be clearer.

nicpottier (approved):
I think this should work! I would REALLY like to see an interface for resolving things instead of passing in two functions though.

search/search.go
rowanseymour: That's fine but note that you can't use an interface - golang doesn't let you implement an interface anonymously 😐

nicpottier: Not sure I understand. Yes, we'd need to create a struct, but then OrgAssets could just satisfy that as would a test version.

rowanseymour: That's what I'm doing - just meant you need a struct.. can't just define the funcs on the fly like we're doing currently

nicpottier: Oh, well I don't much love on the fly funcs anyways so that's fine by me!

rowanseymour:
@nicpottier wanna take another peek?

nicpottier (approved):

PR #245, opened 2020-02-25
Search endpoint should also return whether query can be used as group

rowanseymour:


nicpottier (approved):
Feels like this belongs farther down than the web layer.. either `models/groups.go` or `search`?

rowanseymour:
Yeah - tho I left it there for the time being cuz I think now that we're not doing checks like `"id" in fields` on the RP side, we should replace `fields = ["name", "age"]` with something more like actual dependency list on flows, since I think that's now only used for saving off dynamic group dependencies on contact fields.

PR #250, opened 2020-03-20
TwiML Say language

koallann:
This PR adds the language attribute to the TwiML Say action implementation, based on org primary language.

Ref: https://www.twilio.com/docs/voice/twiml/say?code-sample=code-using-attributes-in-a-say-verb-4&code-language=PHP&code-sdk-version=5.x


web/ivr/ivr.go
```diff
@@ -164,7 +164,7 @@ func handleIncomingCall(ctx context.Context, s *web.Server, r *http.Request, raw
 		resumeURL := buildResumeURL(channel, conn, urn)
 
 		// have our client output our session status
-		err = client.WriteSessionResponse(session, resumeURL, r, w)
+		err = client.WriteSessionResponse(session, resumeURL, org.Org().DefaultLanguage(), r, w)
```
nicpottier: Don't we want this to be the contact language not the org language?

nicpottier (changes requested):
Woot, thanks for this. Looks reasonable but I think we want the contact language not the org language.

web/ivr/ivr.go
koallann: Yes, it looks better, but is the contact language optional? How about trying the contact language, if not set, trying the org language?

nicpottier: I guess actually if we are thinking about what language should be used to speak to the user then that is more dependent on the language that is supported in the flow. IE, even if the contact's language is `spa` if the flow is authored in `eng` we should be reading in `en-US`.

Does that sound right @rowanseymour ?

rowanseymour: It only makes sense to use the language of the text we're reading.. which we calculate from the contact, the flow base language, allowed org languages, and the available translations... and may differ between different parts of the flow (some might be translated into the contact's language, some not).

But at this point we just have a string.. need to think of a clever way of figuring out which translation it came from.

I just made https://github.com/nyaruka/goflow/pull/886 as a possible utility here.

rowanseymour: We need something on the `ivr_created` event for the engine to store which language/translation it picked

rowanseymour: FYI this is now implemented - `ivr_created` events include `textLanguage envs.Language`

You can also call `ToISO639_2` on that to get a 2-char code.. but the tricky part here would still be mapping that to a particular language that Twilio supports

koallann: Great!

koallann: @nicpottier @rowanseymour FYI the changes based on `ivr_created` event were made. Thanks a lot for your attention in these points of IVR!

nicpottier (approved):

nicpottier:
Thanks!

PR #251, opened 2020-03-20
add /org/[key]/metrics endpoint

nicpottier:
Adds prometheus endpoint for contact group counts and channel counts.


web/org/metrics.go
```diff
@@ -0,0 +1,334 @@
...
+	"github.com/pkg/errors"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+func init() {
+	web.RegisterRoute(http.MethodGet, "/mr/org/{token:[0-9a-f]+}/metrics", handleMetrics)
+}
+
+const groupCountsSQL = `
```
rowanseymour, line 22: I think ideally this stuff would live in `models` instead of `web/..`

web/org/metrics.go
```diff
@@ -0,0 +1,334 @@
...
+
+func calculateGroupCounts(ctx context.Context, s *web.Server, org *org) (*dto.MetricFamily, error) {
+	rows, err := s.DB.QueryxContext(ctx, groupCountsSQL, org.ID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error querying group counts for org")
+	}
+	defer rows.Close()
+
+	family := &dto.MetricFamily{
+		Name:   proto.String("rapidpro_group_contact_count"),
```
rowanseymour, line 58: temba? in case textit users want this?

rowanseymour (approved):

web/org/metrics.go
nicpottier: I considered that but it is pretty specific to metrics. Everything in `models..` deals with a bit more concrete stuff. I don't know, just seemed awash either way and having it closer to the point of use felt right since these are highly optimized for this use case.

web/org/metrics.go
nicpottier: So was thinking of using `temba_` for the internal metrics (things like courier queue size for example). `rapidpro_` instead being "org" level stuff. 

We don't need to hide that TextIt is RapidPro I don't think.

web/org/metrics.go
nicpottier: (could I guess obfuscate with `rp_`?)

web/org/metrics.go
rowanseymour: ok that works 👍

web/org/metrics.go
nicpottier: Meh, in the end I think just being clear here is best.

PR #255, opened 2020-03-30
⚗️ Convert all web tests to use same runner

rowanseymour:
Started adding PO export/import endpoint, then started tweaking test stuff to let me have expected output files that aren't JSON, then ended up doing this...


web/testing.go
```diff
@@ -112,22 +41,28 @@ func RunWebTests(t *testing.T, truthFile string) {
 		Label        string          `json:"label"`
 		Method       string          `json:"method"`
 		Path         string          `json:"path"`
-		Body         json.RawMessage `json:"body"`
+		Body         json.RawMessage `json:"body,omitempty"`
 		Status       int             `json:"status"`
-		Response     json.RawMessage `json:"response"`
+		Response     json.RawMessage `json:"response,omitempty"`
+		ResponseFile string          `json:"response_file,omitempty"`
```
rowanseymour, line 47: to be used for PO file responses which will be external to the main JSON file of tests

nicpottier (approved):
one test to rule them

PR #256, opened 2020-03-31
📦 Import and export of POs

rowanseymour:


web/testing.go
```diff
@@ -38,13 +41,14 @@ func RunWebTests(t *testing.T, truthFile string) {
...
-		Path         string          `json:"path"`
-		Body         json.RawMessage `json:"body,omitempty"`
-		Status       int             `json:"status"`
-		Response     json.RawMessage `json:"response,omitempty"`
-		ResponseFile string          `json:"response_file,omitempty"`
+		Label        string            `json:"label"`
+		Method       string            `json:"method"`
+		Path         string            `json:"path"`
+		Body         json.RawMessage   `json:"body,omitempty"`
+		Files        map[string]string `json:"files,omitempty"`
```
rowanseymour, line 48: same as python requests library, if you include `files`, it becomes a multipart request

web/po/po.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return err, http.StatusBadRequest, nil
+	}
+
+	err = i18n.ImportIntoFlows(po, form.Language, flows...)
+	if err != nil {
+		return err, http.StatusBadRequest, nil
+	}
+
+	return flows, http.StatusOK, nil
```
rowanseymour: so this endpoint could make the actual changes to the flows in the database, save a new revision etc.. but it seems like that's best kept as RapidPro's responsibility.. so just returning the new flow definitions here

nicpottier (approved):
PR 256!

PR #261, opened 2020-04-08
Update testing database

rowanseymour:


web/po/testdata/favorites.po
```diff
@@ -24,7 +24,7 @@ msgstr ""
 msgid "Cyan"
 msgstr ""
 
-#: Favorites/4809186b-3f6b-445a-bb81-2f6eda021bc6/text:0
```
rowanseymour, line 27: This could get annoying - everytime we rebuild the mailroom db, the imported flows get new random UUIDs

Might try to find a way to mock that so its the same sequence every time

nicpottier (approved):

PR #262, opened 2020-04-09
Add method to set modified_by on contacts

norkans7:


web/contact/contact.go
```diff
@@ -300,12 +301,23 @@ func handleModify(ctx context.Context, s *web.Server, r *http.Request) (interfac
...
 		}
+		if len(results[scene.ContactID()].Events) > 0 {
+			modifiedContactIDs = append(modifiedContactIDs, scene.ContactID())
+		}
+	}
+
+	// apply modified_by
+	err = models.UpdateContactModifiedBy(ctx, tx, modifiedContactIDs, request.UserID)
+	if err != nil {
+		return nil, http.StatusInternalServerError, errors.Wrapf(err, "error applying modified_by")
```
nicpottier: Probably should be after the pre commit hooks below.

nicpottier (approved):
Small tweak but otherwise looks good.

nicpottier:
So I think this makes sense to do but I wonder whether we should reconsider how we log these types of things. `modified_by` on Contact really is pretty inaccurate, as a flow may have modified the contact since that was last set, so it may be pointing you in the wrong direction entirely.

Wonder whether the right solution here would rather be to create an audit trail for these events that are presented in the contact history the same as flow events are. That would give you a lot more visibility into what is happening and in truth the logging could be very similar since these modifiers also generate events.

But do think this is fine for the time being. @rowanseymour

rowanseymour:
Hmm yeah could have a user field on events when they were initiated from the UI. 

Similar https://github.com/rapidpro/rapidpro/issues/1016

Also fine with this for now

norkans7:
I think it might be useful to have both a user and a flow so the UI could show for each modification what made it a user or a flow or import too

Not sure if we want to have distinction such changed by import or via API by a user

PR #263, opened 2020-04-09
Update flow start modified_on when making changes

rowanseymour:


models/starts.go
```diff
@@ -250,8 +250,8 @@ func InsertFlowStarts(ctx context.Context, db Queryer, starts []*FlowStart) erro
 
 const insertStartSQL = `
 INSERT INTO
-	flows_flowstart(created_on,  uuid,  restart_participants,  include_active,  query, status,  flow_id,  extra,  parent_summary)
-			 VALUES(NOW()     , :uuid, :restart_participants, :include_active, :query, 'P'   , :flow_id, :extra, :parent_summary)
+	flows_flowstart(uuid,  created_on,  modified_on,  restart_participants,  include_active,  query, status,  flow_id,  extra,  parent_summary)
```
rowanseymour, line 253: also need to make this field non-null as it's the field we currently sort by in the API

nicpottier (approved):

PR #265, opened 2020-04-10
Set org ID on new flow starts

rowanseymour:


ivr/ivr.go
```diff
@@ -329,7 +329,7 @@ func StartIVRFlow(
 	}
 
 	// get the flow for our start
-	start, err := models.GetFlowStartAttributes(ctx, db, org.OrgID(), startID)
```
rowanseymour, line 332: just noticed this arg isn't used..

nicpottier (approved):

PR #266, opened 2020-04-10
Update test database to version generated with seeded UUIDs

rowanseymour:


nicpottier (approved):

models/flows_test.go
```diff
@@ -13,15 +14,18 @@ func TestFlows(t *testing.T) {
 	ctx := testsuite.CTX()
 	db := testsuite.DB()
 
+	db.MustExec(`UPDATE flows_flow SET metadata = '{"ivr_retry": 30}'::json WHERE id = $1`, IVRFlowID)
```
rowanseymour, line 17: @nicpottier this is how i stumbled across that `ivr_retry` issue - I needed to update the flows to v13 because the legacy import code isn't deterministic when it comes to handing out new UUIDs - but then this was failing because new flows don't keep things like `expires` in flow metadata

PR #268, opened 2020-04-16
🧯 Don't blow up if flow is deleted during simulation

rowanseymour:
Should fix https://sentry.io/organizations/nyaruka/issues/1610053073/?project=1281372&referrer=alert_email

### DIFF
--- a/web/simulation/simulation.go
+++ b/web/simulation/simulation.go
@@ -225,7 +225,7 @@ func handleResume(ctx context.Context, s *web.Server, r *http.Request) (interfac
 			var flow *models.Flow
 			for _, r := range session.Runs() {
 				if r.Status() == flows.RunStatusWaiting {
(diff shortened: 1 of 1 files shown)


web/simulation/simulation.go
```diff
@@ -225,7 +225,7 @@ func handleResume(ctx context.Context, s *web.Server, r *http.Request) (interfac
 			var flow *models.Flow
 			for _, r := range session.Runs() {
 				if r.Status() == flows.RunStatusWaiting {
-					f, _ := org.Flow(r.Flow().UUID())
+					f, _ := org.Flow(r.FlowReference().UUID)
```
rowanseymour, line 228: `Flow()` can be nil but `FlowReference()` will never let you down

nicpottier (approved):

PR #270, opened 2020-04-20
Add blocked and stopped hooks

norkans7:


hooks/hooks_test.go
```diff
@@ -225,3 +231,61 @@ func RunActionTestCases(t *testing.T, tcs []HookTestCase) {
 		rc.Close()
 	}
 }
+
+func RunEventsTestCases(t *testing.T, tcs []HookEventsTestCase) {
```
norkans7: Added a way to test events are applied even if the they are not flow actions

rowanseymour: Should this be testing modifiers instead? Let them create the events..

norkans7: modifiers are as well tested in 

https://github.com/nyaruka/mailroom/pull/270/files#diff-cc2d4d07b0c749193d82b461cf4403d8

this was to make sure we have coverage for the hooks

rowanseymour: Sure tho you could make the same argument for the other hook tests which are action based - actions are tested well elsewhere, but by testing via actions, we improve our end to end testing.

For bonus points I'd also merge these two testing mechanisms - i.e. one `HookTestCase` that can take actions or modifiers.

norkans7: Okay, I'll update those

norkans7: That is updated

norkans7:
Replaced by https://github.com/nyaruka/mailroom/pull/277

PR #299, opened 2020-06-15
Update to latest goflow v0.92.0

rowanseymour:
Elasticsearch stuff lives in goflow now


web/contact/contact.go
```diff
@@ -44,19 +45,26 @@ type searchRequest struct {
...
-	Sort         string             `json:"sort"`
+	Query      string                `json:"query"`
+	ContactIDs []models.ContactID    `json:"contact_ids"`
+	Total      int64                 `json:"total"`
+	Offset     int                   `json:"offset"`
+	Sort       string                `json:"sort"`
+	Metadata   *contactql.Inspection `json:"metadata,omitempty"`
+
+	// deprecated
+	Fields       []string `json:"fields"`
```
rowanseymour, line 66: i want to replace this with metadata above as it's always been a bit confusing to have something called fields that includes things like URN schemes - and we end up having to filter those out on the RP side

nicpottier (approved):

PR #301, opened 2020-06-17
💳 Make topups optional

rowanseymour:


models/orgs.go
```diff
@@ -60,13 +61,23 @@ const (
 
 // Org is mailroom's type for RapidPro orgs. It also implements the envs.Environment interface for GoFlow
 type Org struct {
-	id     OrgID
-	env    envs.Environment
-	config map[string]interface{}
+	o struct {
+		ID         OrgID    `json:"id"`
+		Suspended  bool     `json:"is_suspended"`
```
rowanseymour, line 66: added this field too but not sure how we would want to use it in mailroom... am thinking courier won't even create messages for orgs that have this set.. yeah? And we'll block a lot of things at a UI level in RapidPro... so is there anything here that actually needs blocked by this being set? Should things that are blocked at UI level in RP be blocked again here in the name of thoroughness?

hooks/ivr_created.go
```diff
@@ -22,7 +22,7 @@ type CommitIVRHook struct{}
 var commitIVRHook = &CommitIVRHook{}
 
 // Apply takes care of inserting all the messages in the passed in scene assigning topups to them as needed.
-func (h *CommitIVRHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, org *models.OrgAssets, scenes map[*models.Scene][]interface{}) error {
+func (h *CommitIVRHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, oa *models.OrgAssets, scenes map[*models.Scene][]interface{}) error {
```
rowanseymour, line 25: i always get confused about what's an `Org` and what's an `OrgAssets`

models/orgs.go
nicpottier: I think the behavior for suspended is the same as the behavior for no topups. You still receive incoming messages etc, we just don't reach to them. So ya, we do want to read it and react accordingly.

hooks/ivr_created.go
nicpottier: Fine with this but maybe these kinds of renames should be done in separate PRs, just because it is a lot more likely I'm going to miss something trying to review when there are a hundred benign renames.

models/orgs.go
nicpottier: Courier doesn't know about topups / plans etc.. and we really don't want to introduce that.

Looking at the current mailroom code it looks like we do handle incoming messages without a topup, just that when the outgoing messages are created we don't assign a topup and the message isn't sent. That doesn't seem ideal.

Seems like we should maybe just not handle. But wonder if we should have a different msg status to represent that otherwise it may get confusing as to why an incoming message wasn't handled.

Should we think about adding a new status to messages that represents the message is failed because of non-payment? We kind of have the same problem with outgoing messages.. as in they are created in a `PENDING` state but it sure won't be clear what occurred there. Alternatively could use `FAILED` but have another field to explain why it failed that is nullable?

nicpottier (approved):

models/orgs.go
rowanseymour: To be clear wasn't suggesting that Courier would know about topups / plans etc.. just that `biz` can set `org.is_suspended` to tell Courier start rejecting messages for an org.

Regarding whether `credits = 0` is the same as `suspended = true`... maybe it should be that there's a grace period between them. You hit `credits = 0` and we still create incoming messages... but if you haven't paid 1 week later.. we stop creating messages completely. Otherwise don't we potentially end up creating messages for dead orgs forever?

rowanseymour: Don't we handle recent messages once an org that had run out of credit adds more credit? I'm not seeing that in the code tho...

nicpottier: I think sometime in the past we may have gone back and re-handled messages but we haven't done that for a while, I think it ended up not being what users actually wanted most of the time.

I think to your question as to what happens with accepting messages I think the answer there is we eventually delete the org entirely so that's how you deal with that. So you are in a suspended / no topup state for a while until we do that.

I *think* it is more intuitive to have messages be created by "failed" in the suspended case as opposed to seeing nothing at all, but maybe I'm wrong there! Suppose if we make put a big badge that you are suspended when you log in maybe not.

Should we slip an org to suspended when they run out of credits (and flip back when they add a topup) so that all the logic is simpler / uniform? Then we could change courier to do the right thing there.

rowanseymour: > Should we slip an org to suspended when they run out of credits (and flip back when they add a topup) so that all the logic is simpler / uniform? 

Yeah I think so - regardless of the rest of this - let's agree that _suspended means an org doesn't have a valid plan_ (it expired or ran out of credits). That can be set by biz or other deployments can set it manually to effectively pause an org without deleting it.

I'm on the fence about whether we should create messages for suspended orgs. On the one hand it feels a little harsh that messages would just stop appearing.. on the other hand.. there would be a big banner at the top of the page saying your org is suspended please pay your bills.

If they do get created, then I think we just make them regular inbox messages.. I think it would get confusing to say that they're failed... and easier to explain to users that we've just switched off flows in their org. 

Agree that after X days of suspension, we should be deleting orgs. That can be in the big warning banner of doom.

nicpottier: I think the argument for creating the messages is just that maybe you aren't "losing" quite as much data then. IE, the messages are still there even if we didn't process them. Never creating them may mean they have no log whatsoever of that message.

Another (small) argument may be that then nothing has to change on the Courier side either if we still create the messages.

tasks/handler/worker.go
```diff
@@ -458,28 +458,37 @@ func handleStopEvent(ctx context.Context, db *sqlx.DB, rp *redis.Pool, event *St
...
 	}
 
+	// if org is suspended, just send message to inbox
+	if oa.Org().Suspended() {
+		err := models.UpdateMessage(ctx, db, event.MsgID, models.MsgStatusHandled, models.VisibilityVisible, models.TypeInbox, models.NilTopupID)
+		if err != nil {
+			return errors.Wrapf(err, "error updating message for suspended org")
+		}
+		return nil
+	}
```
rowanseymour, line 473: @nicpottier added this

PR #303, opened 2020-06-18
⚙️ Handling as normal for suspended orgs

rowanseymour:


hooks/msg_created.go
```diff
@@ -149,12 +149,10 @@ func (h *CommitMessagesHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.P
...
 
-	// find the topup we will assign
-	rc := rp.Get()
-	topup, err := models.AllocateTopups(ctx, tx, rc, oa.Org(), len(msgs))
-	rc.Close()
+	// allocate a topup for this message if org uses topups
+	topup, err := models.AllocateTopups(ctx, tx, rp, oa.Org(), len(msgs))
 	if err != nil {
-		return errors.Wrapf(err, "error finding active topup")
+		return errors.Wrapf(err, "error allocating topup for outgoing message")
```
nicpottier, line 155: Think we also need to change the apply above to queue to courier in the Nil topup case.

nicpottier (changes requested):

hooks/msg_created.go
rowanseymour: Changed in https://github.com/nyaruka/mailroom/pull/303/commits/f16d0c4ada56acd9d32de91e12c9d4c62a81acec

nicpottier (approved):

PR #309, opened 2020-06-26
Update to latest goflow and add tests for field modifiers

rowanseymour:


web/server.go
```diff
@@ -126,7 +127,7 @@ func (s *Server) WrapJSONHandler(handler JSONHandler) http.HandlerFunc {
 			}
 		}
 
-		serialized, serr := json.MarshalIndent(value, "", "  ")
+		serialized, serr := jsonx.MarshalPretty(value)
```
rowanseymour, line 130: uses a decoder which doesn't do the unnecessary encoding of html chars like `>`

norkans7 (approved):

PR #315, opened 2020-07-02
🧯 Fix reading of modifiers again

rowanseymour:
When surveyor reads modifiers it's ok to ignore missing assets and apply what we can can as the surveyor session may be saved some time after it was collected. If we're calling the contact/modify endpoint with missing assets that should be an error because we're doing something wrong.

### DIFF
--- a/goflow/modifiers.go
+++ b/goflow/modifiers.go
@@ -1,0 +1,29 @@
+package goflow
+
+import (
+	"encoding/json"
+
+	"github.com/nyaruka/goflow/assets"
+	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/flows/actions/modifiers"
+
+	"github.com/pkg/errors"
+)
+
+// ReadModifiers reads modifiers from the given JSON
+func ReadModifiers(sa flows.SessionAssets, data []json.RawMessage, allowMissing bool) ([]flows.Modifier, error) {
+	mods := make([]flows.Modifier, 0, len(data))
+	for _, m := range data {
+		mod, err := modifiers.ReadModifier(sa, m, assets.IgnoreMissing)
+
+		// if this modifier turned into a no-op, ignore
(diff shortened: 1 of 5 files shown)


web/contact/contact_test.go
```diff
@@ -194,5 +194,10 @@ func TestModifyContacts(t *testing.T) {
 	db.MustExec(`DELETE FROM contacts_contactgroup_contacts WHERE contact_id = $1`, models.CathyID)
 	db.MustExec(`UPDATE contacts_contacturn SET contact_id = NULL WHERE contact_id = $1`, models.CathyID)
 
+	// because we made changes to a group above, need to make sure we don't use stale org assets
+	models.FlushCache()
```
rowanseymour, line 198: seeing intermittent test failures because of this

goflow/modifiers.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/nyaruka/goflow/assets"
+	"github.com/nyaruka/goflow/flows"
+	"github.com/nyaruka/goflow/flows/actions/modifiers"
+
+	"github.com/pkg/errors"
+)
+
+// ReadModifiers reads modifiers from the given JSON
+func ReadModifiers(sa flows.SessionAssets, data []json.RawMessage, allowMissing bool) ([]flows.Modifier, error) {
```
nicpottier: Super minor style comment but especially as you already have a constant in assets for `IgnoreMissing` maybe worth creating a type for `MissingBehavior` or use constants when calling this function at least. `true`/`false` pretty vague when this is called.

nicpottier (approved):

goflow/modifiers.go
rowanseymour: Updated. It does frustrate me that one has to define new types when what you really want is something as simple as `allow_missing=True` but 🤷 

(`IgnoreMissing` in the other case is actually a callback function)

PR #321, opened 2020-07-21
☔ Add loop protection

rowanseymour:
Implemented for flow actions 

Captura de Pantalla 2020-07-21 a la(s) 16 42 35


hooks/session_triggered.go
```diff
@@ -90,14 +91,20 @@ func (h *InsertStartHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool
 				return errors.Wrapf(err, "error loading contacts by reference")
 			}
 
+			historyJSON, err := jsonx.Marshal(event.History)
+			if err != nil {
+				return err
```
nicpottier: not wrapping an error is basically guaranteeing this will get thrown at some point and we'll wonder where it is coming from. :)

nicpottier (approved):

hooks/session_triggered.go
rowanseymour: that's funny when I wrote this I was thinking.. this can't happen.. it just can't...

rowanseymour: updated it.. don't want to tempt the universe into making this happen

nicpottier: I mean if we really think it is absolutely impossible then we shouldn't even assign to `err` and just use `_`, but ya.. that's even more tempting fate.

rowanseymour: well *if* it did happen it would be because we screwed up at a code level so that's probably an argument for it being a panic.. it's a shame go doesn't have succinct way to let an error go to a panic if it ever happens.. like not listing the error with even `_` on the lefthand could mean panic if it happens.

PR #327, opened 2020-08-03
👀 Update last_seen_on when we receive a message from a contact

rowanseymour:


hooks/contact_modified.go
```diff
@@ -14,17 +15,43 @@ type ContactModifiedHook struct{}
 
 var contactModifiedHook = &ContactModifiedHook{}
 
+type modifiedEvent struct {
+	InputReceived bool
+}
+
 // Apply squashes and updates modified_on on all the contacts passed in
 func (h *ContactModifiedHook) Apply(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, oa *models.OrgAssets, scenes map[*models.Scene][]interface{}) error {
```
rowanseymour, line 23: maybe over-optimized..? But since updating last_seen_on requires updating modified_on.. I've combined them into one hook so we hit the contact table once

nicpottier: Aren't you still hitting it twice though or am I missing something? To me this feels like these should be separate hooks as the whole point of the hooks and callbacks it to make those easier to reason about. They are all batched into transactions either way. You also wouldn't need to do this scene-reading as you are doing here.

nicpottier (changes requested):

hooks/contact_modified.go
rowanseymour: So situations in which these fields get updated..

 1. Inbox message - need to update `last_seen_on` and `modified_on` for single contact
 2. Trigger or resume by msg - need to update `last_seen_on` and `modified_on` for single contact
 3. Non-msg start - need to only update `modified_on` for one or more contacts.

This was just a way of saying that 2. should make one `UPDATE` and use same method as 1.

Does them being in the same transaction negate the need to minimize `UPDATE`s ?

It actually got me thinking that if you really want to optimize things you'd have one hook for updates to the contact table, e.g. something like `contactAttributesHook` which you'd use for `name`, `language`, `timezone`, `modified_on`, `last_seen_on`. In its `Apply` it would be able to combine any combination of them into a single `UPDATE`. Currently they would all happen as separate updates for the same contact.

But maybe that's optimizing the wrong thing...

nicpottier: Oh I see, I didn't dive down into `UpdateContactLastSeenOn` to realize that that was also updating `modified_on`. Ok, this makes sense then, this is a common enough scenario that I think you are right batching it makes sense.

From what I can tell, multiple UPDATEs on the same row do cost more, even in a transaction (though I think much less than if they were outside a transaction). 

Agree that `contactAttributesHook` could make sense as a way of optimizing that further, but I don't think that makes sense unless we actually see that as a problem.. ie, `premature optimization is ..`

nicpottier (approved):

PR #328, opened 2020-08-04
👀 Update last_seen_on when handling certain channel events

rowanseymour:


models/contacts.go
```diff
@@ -921,6 +927,14 @@ WHERE
 	id = $1
 `
 
+// UpdateLastSeenOn updates the last seen on for the contact
+func (c *Contact) UpdateLastSeenOn(ctx context.Context, tx Queryer) error {
+	t := dates.Now()
+	c.lastSeenOn = &t
+	c.modifiedOn = t
```
rowanseymour: pre-empting the need to expose this to the engine and keep changes to the db in sync with a model we've already loaded

tasks/handler/worker.go
```diff
@@ -320,29 +320,42 @@ func HandleChannelEvent(ctx context.Context, db *sqlx.DB, rp *redis.Pool, eventT
...
+
 	switch eventType {
 
 	case models.NewConversationEventType:
 		trigger = models.FindMatchingNewConversationTrigger(oa, channel)
+		updateLastSeen = true
 
 	case models.ReferralEventType:
 		trigger = models.FindMatchingReferralTrigger(oa, channel, event.ExtraValue("referrer_id"))
+		updateLastSeen = true
```
rowanseymour: what does it mean to be seen? @nicpottier @ericnewcomer

tasks/handler/worker.go
```diff
@@ -320,29 +320,42 @@ func HandleChannelEvent(ctx context.Context, db *sqlx.DB, rp *redis.Pool, eventT
...
 
 	case models.MOMissEventType:
 		trigger = models.FindMatchingMissedCallTrigger(oa)
+		updateLastSeen = true
 
 	case models.MOCallEventType:
 		trigger = models.FindMatchingMOCallTrigger(oa, modelContact)
+		updateLastSeen = true
```
rowanseymour: Can you have a call which ends before there is input from the contact?

nicpottier: Sure, an incoming call could just read a message and hang up.

nicpottier (approved):
Are we updating last_seen on incoming messages which are not handled by a flow?

rowanseymour:
@nicpottier yeah that was previous PR see https://github.com/nyaruka/mailroom/blob/master/tasks/handler/worker.go#L642-L653

tasks/handler/worker.go
rowanseymour: Ok so then I think any channel event except the outgoing calls and welcome message count as seen. Even stop events are the contact interacting with us, so fo consistency I think that counts. Same as if contact sent the message STOP and a flow set them to stopped.

nicpottier: Agreed

tasks/handler/worker.go
ericnewcomer: Age old question. You've hit all the ones I can think of.

ericnewcomer (approved):

rowanseymour:
@nicpottier @ericnewcomer tweaked a bit now so includes other event types not in that switch like `stop_contact` and am updating before the contact is loaded - but need to think about how `last_seen_on` might be used in the engine and whether it's correct to defer updating it until `msg_received` handling, i.e. if it's exposed in queries and possibly as an attribute of `@contact` you might expect it to be up to date during a sprint. That doesn't necessarily mean that we need to update it in the database prior to the sprint.. something to think about

nicpottier:
Ya, I think I *would* expect `last_seen_on` to be set to the date of the message when processing a flow. Starts getting a bit weird with `modified_on` (do we expose that?) though. We need `modified_on` to change for indexing purposes but it does fee like maybe that has to be a field that we retain for internal purposes otherwise it is going to get confusing to users that `modified_on` changes based only on an incoming message.

rowanseymour:
Engine has no `modified_on` for contacts and the `modified_on` we set in the database is always `NOW()` and not actually based on anything from the engine. That makes sense since that field is just for consumers like indexer, casepro to pick up new db changes to contacts.

Currently last_seen_on is just being set to `NOW()` so that it could be updated in bulk - but on reflection that's not helpful since if we have input, then we're not processing a batch a contacts. I think I should update that to be the actual time of the `msg_received` or channel event.

Then either when we're loading a contact to give to the engine or in the engine itself we can bump the value of `last_seen_on` if there is input to be the time of that input. We can still defer updating it in the database until handling of `msg_received` so that still happens in the transaction.

nicpottier:
All sounds reasonable. Does bring up the point that CasePro and U-Report both poll contacts based on `modified_on` I think. Are we going to expose `last_seen_on` to the API? Can't decide if we should, does seem useful but it does mean a lot more contacts will be appearing as "changed" in CasePro and U-Report than were previously. (basically every contact participating in a poll for example, which could be problematic for both)

rowanseymour:
I think we should because currently we need to update `modified_on` when we change `last_seen_on` for the benefit of indexer.. so CasePro etc who are polling contacts based on `modified_on` are going to see contacts which have supposedly changed.. but they won't appear any differently... unless we add that field so they can see a difference.

Or we have a new date field on contacts `api_modified_on` which isn't updated for things that won't appear over the API.

PR #329, opened 2020-08-05
⌚ Last seen should be set as event time

rowanseymour:


hooks/contact_last_seen.go
```diff
@@ -0,0 +1,31 @@
...
+	"time"
+
+	"github.com/gomodule/redigo/redis"
+	"github.com/jmoiron/sqlx"
+	"github.com/nyaruka/mailroom/models"
+	"github.com/pkg/errors"
+)
+
+// ContactLastSeenHook is our hook for contact changes that require an update to last_seen_on
+type ContactLastSeenHook struct{}
```
rowanseymour, line 14: now that this isn't setting `NOW()` same as `contactModifiedHook` I can't convince myself they should be the same hook

nicpottier (approved):

PR #333, opened 2020-08-14
RocketChat Ticketer

koallann:
https://github.com/nyaruka/rapidpro/pull/2909


services/tickets/rocketchat/testdata/event_callback.json
```diff
@@ -0,0 +1,143 @@
+[
+  {
+    "label": "error response if no such ticketer",
+    "method": "POST",
+    "path": "/mr/tickets/types/rocketchat/event_callback/XYZ",
+    "body": "{\"type\":\"agent-message\",\"ticketID\":\"88bfa1dc-be33-45c2-b469-294ecb0eba90\",\"visitor\":{\"token\":\"1234\"},\"data\":{\"text\":\"We can help\"}}",
```
rowanseymour: don't need to encode as string here - should work as JSON

rowanseymour:
@koallann this looks good but you still want to add attachment support right? We're planning code freeze for RapidPro in 2 weeks time so would be great to get this included.

koallann:
@rowanseymour almost fully implemented, we only need an [adjustment](https://github.com/rapidpro/rapidpro/issues/1309#issuecomment-709223973) to incoming attachments.

rowanseymour:
@koallann ok maybe just hold off on the incoming attachments thing because I started looking at what that would require for mailgun/email ticketers and that's not going to work with fetching URLs because the attachments are embedded in the request... I'm going to add stuff today to support getting attachments from different places.

But this is ready minus incoming attachments?

services/tickets/rocketchat/web_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/nyaruka/mailroom/web"
+	"testing"
+)
+
+func TestEventCallback(t *testing.T) {
+	testsuite.Reset()
+	db := testsuite.DB()
+
+	// create a rocketchat ticket for Cathy
+	db.MustExec(`INSERT INTO tickets_ticket(id, uuid, org_id, contact_id, ticketer_id, status, subject, body, opened_on, modified_on)
```
rowanseymour: there's now a `testdata.InsertOpenTicket` util function

koallann:
Yes @rowanseymour

rowanseymour:
@koallann ok have just merged #374 which reworks incoming attachments but RocketChat should be basically the same as Zendesk with the addition of passing some headers

koallann:
done @rowanseymour

services/tickets/rocketchat/service.go
```diff
@@ -0,0 +1,151 @@
...
+			ContactUUID: string(contact.UUID()),
+			Name:        contact.Name(),
+			Email:       email,
+			Phone:       phone,
+		},
+		TicketID: string(ticketUUID),
+	}
+	room.SessionStart = session.Runs()[0].CreatedOn().Add(-time.Minute).Format(time.RFC3339)
+
+	// to fully support the RocketChat ticketer, look up extra fields from ticket body for now
```
rowanseymour: I appreciate this is maybe how you guys are using RocketChat right now but I think we need a way of sharing fields etc with external ticketing services which provides a more uniform experience across different ticketing services, i.e. body isn't expected to be JSON for other services, and this approach would mean the ticket list view page becomes a list of big JSON blobs.

It's also not a great approach in that you're sending a snaphot of field values when the ticket is opened but users in RocketChat won't see changes to those field values.

For Zendesk we haven't yet implemented syncing field values but it will probably either be

 1. The Zendesk app can fetch them from the RapidPro API
 2. Every time we push a message to Zendesk we push a set of field values that can be configured on the ticketer in RApidPro

services/tickets/rocketchat/service.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+		if err != nil {
+			return errors.Wrap(err, "error calling RocketChat")
+		}
+	}
+	return nil
+}
+
+func (s *service) Reopen(tickets []*models.Ticket, logHTTP flows.HTTPLogCallback) error {
+	return errors.New("RocketChat ticket type doesn't support reopening")
```
rowanseymour, line 133: Need some way of telling the RapidPro UI to not try re-opening tickets associated with RocketChat ticketers

services/tickets/rocketchat/service.go
rowanseymour: I think maybe we remove this for now, and revisit how to send field values later.

rowanseymour (changes requested):

services/tickets/rocketchat/service.go
koallann: I totally agree with you, but these fields are essential for our customers because each of them has a different context of use. The data in these fields is very important for agents in RocketChat to be able to make the information triage. With this information, the agent can decide to get the conversation or forward it to another department more appropriate to the user/contact, for example.

rowanseymour: Are you already running this as a fork, or is this to replace existing changes you've made to RP to support RocketChat? This feels like a "hack" so we could merge without it and then you can put it back into your fork..

koallann: We're still not running this code, but we can run that hack on a fork, so I'll remove it.

services/tickets/rocketchat/client_test.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"github.com/nyaruka/gocommon/httpx"
+	"github.com/nyaruka/mailroom/services/tickets/rocketchat"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"testing"
+)
+
+const (
+	baseURL = "https://my.rocket.chat/api/apps/public/684202ed-1461-4983-9ea7-fde74b15026c"
```
rowanseymour, line 12: @koallann where can we see the source for this?

koallann: The UUID in this url is the RocketChat app ID and the app API is [here](https://github.com/Ilhasoft/Rocket.Chat.App-Tickets/tree/release/v1.0.0#api-reference).

koallann:
Hey @rowanseymour, any more changes here?

rowanseymour:
@koallann I don't think so tho we're focused on getting the Rocket.Chat channel PRs live and tested first

PR #334, opened 2020-08-14
⌚ Add task to calculate fires for new campaign event

rowanseymour:


models/campaigns.go
```diff
@@ -586,3 +587,129 @@ func AddCampaignEventsForGroupAddition(ctx context.Context, tx Queryer, org *Org
...
+const eligibleContactsForFieldSQL = `
+SELECT 
+	c.id AS contact_id, 
+	c.fields->$2->>'datetime' AS value
+FROM 
+	contacts_contact c
+INNER JOIN
+    contacts_contactgroup_contacts gc ON gc.contact_id = c.id
+WHERE
+    gc.contactgroup_id = $1 AND c.is_active = TRUE AND c.fields->$2->>'datetime' IS NOT NULL
```
nicpottier: I think this is going to be a scan against every contact in an org.. see:
  https://github.com/nyaruka/rapidpro/blob/master/temba/campaigns/models.py#L623

On how to use the dynamic jsonb index we build to efficiently get the contacts with that field value set.

models/campaigns.go
```diff
@@ -586,3 +587,129 @@ func AddCampaignEventsForGroupAddition(ctx context.Context, tx Queryer, org *Org
...
+	case LastSeenOnKey:
+		query = eligibleContactsForLastSeenOnSQL
+		params = []interface{}{groupID}
+	default:
+		query = eligibleContactsForFieldSQL
+		params = []interface{}{groupID, pq.Array([]interface{}{field.UUID()})}
+	}
+
+	rows, err := db.QueryxContext(ctx, query, params...)
+	if err != nil {
```
nicpottier: I think Golang SQL will return `sql.ErrNoRows` if nothing matches, so you want to account for that.

models/campaigns.go
```diff
@@ -586,3 +587,129 @@ func AddCampaignEventsForGroupAddition(ctx context.Context, tx Queryer, org *Org
 	// add all our new event fires
 	return AddEventFires(ctx, tx, fas)
 }
+
+// CalculateCampaignEventFires calculates event fires for a new campaign event
+func CalculateCampaignEventFires(ctx context.Context, db *sqlx.DB, orgID OrgID, eventID CampaignEventID) error {
+	// NOOP if fires already exist for this event
```
nicpottier: Doesn't an edit of a campaign event require deleting old fires and recalculating new ones, or are you going to handle the deletion elsewhere?

nicpottier (changes requested):

models/campaigns.go
rowanseymour: We don't delete fires any more when events are edited - we clone the event and deactivate the old one so when its event fires fire.. we ignore them.

models/campaigns.go
rowanseymour: Ahhhh ok so that's why we use that check on the field keys - that makes sense now!

models/campaigns.go
rowanseymour: doh yes.. tho this is a draft PR with zero tests 🤸

rowanseymour:
@nicpottier wanna put this to bed before moving on to other things - please take another look!

models/campaigns.go
```diff
@@ -586,3 +587,134 @@ func AddCampaignEventsForGroupAddition(ctx context.Context, tx Queryer, org *Org
 	// add all our new event fires
 	return AddEventFires(ctx, tx, fas)
 }
+
+// ScheduleCampaignEvent calculates event fires for a new campaign event
+func ScheduleCampaignEvent(ctx context.Context, db *sqlx.DB, orgID OrgID, eventID CampaignEventID) error {
+	// NOOP if fires already exist for this event
+	var fireID FireID
+	db.Get(&fireID, "SELECT id FROM campaigns_eventfire WHERE event_id = $1", eventID)
```
nicpottier: Doesn't this need a `LIMIT 1`?

nicpottier: And isn't there a bit of a race here that giving up so easily could be wrong? As soon as a user creates a new campaign event we will start assigning contacts to that event in flows if their group membership changes etc.. seems possible that that could occur before this task gets to this point no?

models/campaigns.go
```diff
@@ -586,3 +587,134 @@ func AddCampaignEventsForGroupAddition(ctx context.Context, tx Queryer, org *Org
...
+			}
+
+			if scheduled != nil {
+				fas = append(fas, &FireAdd{ContactID: el.ContactID, EventID: eventID, Scheduled: *scheduled})
+			}
+		}
+	}
+
+	// add all our new event fires
+	return AddEventFires(ctx, db, fas)
```
nicpottier, line 634: This could be a bit much to commit in one go for cases like Nigeria and the like. Probably need to batch this into reasonable chunks.

nicpottier: (that probably belongs in `AddEventFires`)

nicpottier (changes requested):

models/campaigns.go
rowanseymour: I copied this from the Python but I think we can get rid of it since yes there's a potential race, and it seems to only serve as protection against queuing this task twice.. which I'm pretty confident we can avoid doing.

models/campaigns.go
rowanseymour: or in BulkSQL?

models/campaigns.go
rowanseymour: Tho then I think we also need to skip over contacts who already have a fire

models/campaigns.go
nicpottier: Ya, probably fine having it in BulkSQL as well, though having that be an optional behavior.

models/campaigns.go
nicpottier: Should we have a uniqueness constraint on event/fire_date/contact for unfired events?

models/campaigns.go
rowanseymour: what's the downside of having BulkSQL always batch things?

nicpottier: I'm not 100% sure.. all I know is that a lot of mailroom is fairly carefully reasoned SQL from a performance / trigger / conflict perspective and I feel slightly sheepish about changing that across all calls without them opting in. Maybe that's undue paranoia but I feel like if it isn't broken then let's not "fix" those calls. (most places that need batching are already doing them)

One tricky bit to be aware of is that that BulkSQL call also populates fields on the passed in objects when doing inserts etc.. so that behavior has to stay the same.

rowanseymour: That's funny cuz I just made https://github.com/nyaruka/mailroom/pull/352 really for the purposes of better testing the bulk stuff and making sure I understand exactly what it's doing. The returning ids stuff tested too.

My feeling is we can't let ourselves get into situations where we're afraid to touch code.. if we're concerned about `BulkSQL` making SQL statements that are too long in this particular case.. then that's a potential concern anywhere that function is used. Let's make it batch, test that well, and have one less thing to worry about in future.

rowanseymour: Maybe I make it optional for now.. but if it works for campaign fires.. then eventually it becomes the default. It also won't kick in most cases if the batch size is high and we're already batching in other ways in other places.

nicpottier: I mean, I don't disagree with not being afraid to change things in general, but mailroom is super high concurrency and it has a lot of DB optimizations baked in. Unless we have CI that is testing those things at scale I think we SHOULD be paranoid about affecting/changing performance characteristics, especially in such a sweeping way that would affect every call.

rowanseymour: Don't disagree... and I think we can have a healthy amount of paranoia whilst making a change like this - thus I've spent a few hours making sure that code is now better unit tested. We know giant SQL commits are a bad idea so let's try batching with event fires.. and if we don't see any problems.. introduce it elsewhere.

rowanseymour: Update to use new `BulkQueryBatches` function (see https://github.com/nyaruka/mailroom/pull/334/files#diff-704543679807b791ba561b05c68badc2R56-R70)

nicpottier (approved):

models/campaigns.go
rowanseymour: Hmm this is starting to sound like something that leads to deadlocks.. tho not really sure 😩

Could the constraint be event_id + contact ? Should you ever have more than one unfired fire for the same event?

nicpottier: An event can fire more than once for a contact.. ie, any repeating campaign.

rowanseymour: Yes but unfired fires..

nicpottier: Confused, isn't that what I said? IE, event/contact where fired is null?

rowanseymour: > event/fire_date/contact for unfired events?

Ok I thought by `fire_date` you meant the scheduled date.. so we're saying `event_id` + `contact_id` should be unique where `fired` is `NULL`

Postgres doesn't actually give us a way to define a partial uniqueness constraint.. and if we did a unique constraint across `event_id` + `contact_id` + `fired` it would consider each NULL to be unique anyways.. but..

I think we can do a partial unique index like https://www.enterprisedb.com/postgres-tutorials/postgresql-unique-constraint-null-allowing-only-one-null

rowanseymour: Alternatively.. you don't try to restrict creating event fires (still concerned about deadlocks doing this), and instead when firing events you ignore a fire if the contact has a newer one for the same event.

nicpottier: I can't think of any way of doing this that isn't a race of some kind.. by definition the two events would have the same fire date so have a high likelihood of happening concurrently. We also display these in the UI so I think having two in the DB could get confusing to users though I suppose we could filter there.

What's the concern from a deadlock perspective? That it is locking the contact table or something more than it would otherwise? I don't think that lock is any different than what it already does to maintain the FK constraint is it?

rowanseymour: Locking the event fire table if we have campaign event re-evaluations happening inside the event committing transactions as well as here... but I really don't have a clear picture of how that would happen so let's proceed with the unique constraint idea...

So I think we create an index like..

```
CREATE UNIQUE INDEX campaigns_eventfire_unique 
ON campaigns_eventfire (event_id, contact_id, (fired IS NULL)) 
WHERE fired IS NULL;
```

And then include an `ON CONFLICT` clause on the add fires SQL to do nothing in case of a conflict

PR #340, opened 2020-08-21
Fix inserting channel logs and add test

rowanseymour:
Fix https://sentry.io/organizations/nyaruka/issues/1851453463/?project=1281372&referrer=alert_email


rowanseymour:
Note that IVR code swallows the error from failing to save the channel log so existing tests didn't catch this

PR #342, opened 2020-08-24
🗽 Read from contact.status instead of is_stopped/is_blocked

rowanseymour:


tasks/handler/worker.go
```diff
@@ -218,8 +218,8 @@ func handleTimedEvent(ctx context.Context, db *sqlx.DB, rp *redis.Pool, eventTyp
 		return errors.Wrapf(err, "error loading contact")
 	}
 
-	// contact has been deleted or is blocked, ignore this event
-	if len(contacts) == 0 || contacts[0].IsBlocked() {
+	// contact has been deleted or is blocked/stopped/archived, ignore this event
+	if len(contacts) == 0 || contacts[0].Status() != models.ContactStatusActive {
```
rowanseymour, line 222: so technically a change of behaviour.. but seems stopped contacts shouldn't get timed events same as blocked?

models/contacts.go
```diff
@@ -52,10 +52,18 @@ const (
...
+	flows.ContactStatusBlocked:  ContactStatusBlocked,
+	flows.ContactStatusStopped:  ContactStatusStopped,
+	flows.ContactStatusArchived: ContactStatusArchived,
+}
+
+var contactToFlowStatus = map[ContactStatus]flows.ContactStatus{
+	ContactStatusActive:   flows.ContactStatusActive,
+	ContactStatusBlocked:  flows.ContactStatusBlocked,
+	ContactStatusStopped:  flows.ContactStatusStopped,
+	ContactStatusArchived: flows.ContactStatusArchived,
```
rowanseymour, line 66: writing this had me wondering if single char codes were worth it in the db.. but then writing https://github.com/nyaruka/rapidpro/pull/2926/files#diff-3e440a4651b73b0746e52e4bcf4376f4R30-R31 convinced it was fine to be consistent for now

PR #347, opened 2020-09-02
📇 Use status for elastic queries that need to filter out non-active contacts

rowanseymour:


models/contacts.go
```diff
@@ -209,7 +209,7 @@ func ContactIDsFromReferences(ctx context.Context, tx Queryer, org *OrgAssets, r
 }
 
 // BuildElasticQuery turns the passed in contact ql query into an elastic query
-func BuildElasticQuery(org *OrgAssets, group assets.GroupUUID, excludeIDs []ContactID, query *contactql.ContactQuery) elastic.Query {
+func BuildElasticQuery(org *OrgAssets, group assets.GroupUUID, activeOnly bool, excludeIDs []ContactID, query *contactql.ContactQuery) elastic.Query {
```
nicpottier: Wonder if should this just be a status passed in?

nicpottier (approved):

models/contacts.go
rowanseymour: It can be.. if you want to get rid of using a bool. But long term... I think you have `status = active` in the actual query 😎

nicpottier: You know how I feel about `bool`s as parameters but I'm down for either.

PR #348, opened 2020-09-04
Empty contact names and languages should be saved as NULL

rowanseymour:
Or at least this is how Django has always done it. Noticed failing tests related to this when switching mailroom_db to use mailroom-side contact creation.


nicpottier (approved):
Eek, ya for sure.

PR #349, opened 2020-09-07
Fix and simplify creation of channel logs in IVR handlers

rowanseymour:
Keep trying and failing to come up with a generic pattern for logging all requests - problem is handler wrapper functions don't mix and different handlers need to resolve different things in order to create logs - e.g. ticketer, channel etc. But this is a start and simplifies the creating channel logs for IVR handlers.


rowanseymour:
Another way that might be cleaner is implementing stuff like this as actual middleware so that it doesn't require changing handler signatures. Handler functions would need to use the context to communicate back things like the channel or ticketer and then the middleware would create and save the logs.

rowanseymour:
This also fixes that currently we don't properly log responses to IVR requests to mailroom as we don't log headers - only the body

PR #350, opened 2020-09-07
🎨 Fix incoming attachments from Zendesk

rowanseymour:


models/orgs.go
```diff
@@ -174,6 +175,7 @@ func (o *Org) StoreAttachment(s storage.Storage, filename string, content []byte
 	prefix := config.Mailroom.S3MediaPrefix
 
 	contentType := http.DetectContentType(content)
+	contentType, _, _ = mime.ParseMediaType(contentType)
```
rowanseymour, line 178: strips off extra stuff like encoding

PR #357, opened 2020-09-22
🧹 Pass orgID from the base task to task structs 

rowanseymour:
To remove need for duplicating it in the task "body" - some of our tasks have ended up with org being included twice in their overall payload. I think this might have been me forgetting that I could get org from the container task so some of the tasks look like...

```json
{
  "type": "schedule_campaign_event",
  "org_id": 123,
  "task": {"org_id": 123, "campaign_event_id": 3456},
  "queued_on": "2020-09-22T10:15:30Z",
}
```

### DIFF
--- a/tasks/base.go
+++ b/tasks/base.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/nyaruka/goflow/utils"
 	"github.com/nyaruka/mailroom"
+	"github.com/nyaruka/mailroom/models"
 	"github.com/nyaruka/mailroom/queue"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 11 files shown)


nicpottier (approved):
I think for a bit I was trying to keep the "org id" as something more generic with the thought that it could be "fair queuing" on something apart from org (such as channel in the courier case) but ya, seems like in mailroom world it is always org id.

rowanseymour:
@nicpottier something has gone wild in the coverage which either means codecov getting confused or trying to tell us that something huge has broken... it really seems like the former.. but I'll proceed with caution

PR #358, opened 2020-09-23
🚀 Imports 2.0

rowanseymour:


models/contacts.go
```diff
@@ -640,58 +640,53 @@ func CreateContact(ctx context.Context, db *sqlx.DB, org *OrgAssets, userID User
...
-
-	contactID, err := insertContactAndURNs(ctx, db, org, UserID(1), "", envs.NilLanguage, []urns.URN{urn})
+// GetOrCreateContact fetches or creates a new contact for the passed in org with the passed in URNs.
+//
+// * If none of the URNs exist, it creates a new contact with those URNs.
+// * If URNs exist but are orphaned it creates a new contact and assigns those URNs to them.
+// * If URNs exists and belongs to a single contact it returns that contact (other URNs are not assigned to the contact).
+// * If URNs exists and belongs to multiple contacts it will return an error.
+//
+func GetOrCreateContact(ctx context.Context, db *sqlx.DB, oa *OrgAssets, urnz []urns.URN) (*Contact, *flows.Contact, error) {
```
rowanseymour: Contract is a bit different than what we've had in RP - we don't promise that the returned contact owns all of the passed in URNs - instead the import code assigns any missing URNs using modifiers. I like this because:

 - Unless we added locking - that stuff isn't going to be atomic anyway
 - We don't duplicate the logic around creating new URNs vs attaching orphaned URNs in that modifier
 - Using modifiers means there's an event to say that a URN was added to a contact... and one day we'll show all events for all contacts... one day

models/contacts.go
```diff
@@ -557,141 +583,128 @@ WHERE
...
-				if !found {
-					return nil, errors.Wrapf(err, "unable to find original URN from identity")
-				}
-				urnMap[original] = contact.ID()
-			}
-		}
+// CreateContact creates a new contact for the passed in org with the passed in URNs
+func CreateContact(ctx context.Context, db *sqlx.DB, oa *OrgAssets, userID UserID, name string, language envs.Language, urnz []urns.URN) (*Contact, *flows.Contact, error) {
+	// find current owners of these URNs
+	owners, err := contactIDsFromURNs(ctx, db, oa.OrgID(), urnz)
```
rowanseymour, line 652: same check to see if URNs are taken added here to avoid even more tx rollbacks

models/contacts.go
```diff
@@ -700,78 +713,126 @@ func GetOrCreateContact(ctx context.Context, db *sqlx.DB, org *OrgAssets, urn ur
...
+	} else if len(uniqueOwners) == 1 {
+		return uniqueOwners[0], false, nil
+	}
+
+	contactID, err := tryInsertContactAndURNs(ctx, db, orgID, UserID(1), "", envs.NilLanguage, urnz)
+	if err == nil {
+		return contactID, true, nil
+	}
+
+	if dbutil.IsUniqueViolation(err) {
```
rowanseymour, line 750: thinking of ways to test this... could have another interface like `Queryer` that supports transactions (`QueryerWithTx`).. use that everywhere instead of `*sqlx.DB`, and make an implementation for testing that we can configure to blow up when needed

models/contacts.go
rowanseymour: see https://github.com/nyaruka/mailroom/pull/360

models/contacts.go
```diff
@@ -183,29 +197,41 @@ func GetNewestContactModifiedOn(ctx context.Context, db *sqlx.DB, org *OrgAssets
...
+	return ids, nil
+}
+
+// utility to query contact IDs
+func queryContactIDs(ctx context.Context, tx Queryer, query string, args ...interface{}) ([]ContactID, error) {
+	ids := make([]ContactID, 0, 10)
+	rows, err := tx.QueryxContext(ctx, query, args...)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error selecting contact ids by uuid")
+		return nil, err
```
nicpottier: Remember go annoyingly treats no rows as an error.

models/contacts.go
```diff
@@ -183,29 +197,41 @@ func GetNewestContactModifiedOn(ctx context.Context, db *sqlx.DB, org *OrgAssets
...
 	}
 	defer rows.Close()
 
 	var id ContactID
 	for rows.Next() {
 		err := rows.Scan(&id)
 		if err != nil {
-			return nil, errors.Wrapf(err, "error scanning contact id")
+			return nil, err
```
nicpottier: What's your rule for wrapping or not wrapping? I kinda just defer to always wrapping because hey, more info when looking at a trace never seems like a bad thing but totally admit maybe that's overkill. Just wondering what rule you use.

models/contacts.go
nicpottier: Don't know if this is the best, or even a good way, but in Courier we do in goroutines by having a flag we can set on the test to slow things down:
  https://github.com/nyaruka/courier/blob/master/backends/rapidpro/backend_test.go#L254
  https://github.com/nyaruka/courier/blob/master/backends/rapidpro/contact.go#L180

models/imports.go
```diff
@@ -0,0 +1,319 @@
...
+// ContactImportBatchID is the type for contact import batch IDs
+type ContactImportBatchID int64
+
+// ContactImportStatus is the status of an import
+type ContactImportStatus string
+
+// import status constants
+const (
+	ContactImportStatusPending    ContactImportStatus = "P"
+	ContactImportStatusProcessing ContactImportStatus = "O"
```
nicpottier, line 34: Would `Running` be better maybe?

models/imports.go
```diff
@@ -0,0 +1,319 @@
...
+	NumErrored int             `db:"num_errored"`
+	Errors     json.RawMessage `db:"errors"`
+	FinishedOn *time.Time      `db:"finished_on"`
+}
+
+// Import does the actual import of this batch
+func (b *ContactImportBatch) Import(ctx context.Context, db *sqlx.DB, orgID OrgID) error {
+	// if any error occurs this batch should be marked as failed
+	if err := b.tryImport(ctx, db, orgID); err != nil {
+		b.markFailed(ctx, db)
```
nicpottier, line 62: Any benefit to logging the error in the task itself too?

models/imports.go
```diff
@@ -0,0 +1,319 @@
...
+	modifiersByContact := make(map[*flows.Contact][]flows.Modifier, len(imports))
+	for _, imp := range imports {
+		// ignore errored imports which couldn't get/create a contact
+		if imp.contact != nil {
+			modifiersByContact[imp.flowContact] = imp.mods
+		}
+	}
+
+	// and apply in bulk
+	_, err = ApplyModifiers(ctx, db, nil, oa, modifiersByContact)
```
nicpottier, line 116: May not be worth it but one thing we do in batch starts is if we fail we retry in individual transactions for each contact.. this helps with deadlocks.

models/imports.go
```diff
@@ -0,0 +1,319 @@
...
+	now := dates.Now()
+	b.Status = ContactImportStatusComplete
+	b.NumCreated = numCreated
+	b.NumUpdated = numUpdated
+	b.NumErrored = numErrored
+	b.Errors = errorsJSON
+	b.FinishedOn = &now
+	_, err = db.ExecContext(ctx,
+		`UPDATE contacts_contactimportbatch SET status = $2, num_created = $3, num_updated = $4, num_errored = $5, errors = $6, finished_on = $7 WHERE id = $1`,
+		b.ID, b.Status, b.NumCreated, b.NumUpdated, b.NumErrored, b.Errors, b.FinishedOn,
```
nicpottier: Since you have a batch at this point maybe cleaner / less error prone to use NamedQuery?

nicpottier (approved):

models/imports.go
rowanseymour: Maybe but these are the same codes that we use in the 3 export models

models/imports.go
rowanseymour: yeah will update

models/contacts.go
rowanseymour: So the code that it replaces wasn't checking for `sql.NoRows` because it's only only sued in places where you're expecting get back ids for things that should exist.. so I went back and forth on whether this being a generic utility method should allow that.. and couldn't make up my mind.. but this morning I'm leaning toward it allowing no rows, and the calling code should check that it got as many ids as it expected

models/contacts.go
rowanseymour: I've generally leaned toward always wrap - not even sure why I didn't here - fine with more wrapping

models/contacts.go
rowanseymour: Interesting - I like the idea of being able to test an actual race - tho I think I'd still want to keep that out of the production code. I.e. could use the db interface as per https://github.com/nyaruka/mailroom/pull/360 but make one of those calls take an extra second.

models/imports.go
rowanseymour: I think this is an error for us (it's a a bunch of wrapped err's) rather than the user facing errors what we collect in `batch.errors` and relate to problems in the user's data, so it's fine if it just goes to sentry.

models/contacts.go
rowanseymour: Got this now https://github.com/nyaruka/mailroom/pull/358/files#diff-87ee1a97bf30c9671e799d60a8bbd25cR663

models/contacts.go
nicpottier: Nice

models/imports.go
rowanseymour: So that would be something like reworking `HandleAndCommitEvents` to catch any error when committing the transaction that wraps `ApplyEventPreCommitHooks`.. and if it fails, try each scene one by one? That sounds pretty straightforward.. tho inclined to try without first.

PR #360, opened 2020-09-29
🧪 Add mockable DB to enable testing database errors

rowanseymour:
Mostly want this for testing race condition stuff around contact creation, but prototyping it here for a simple db query error test

### DIFF
--- a/models/groups_test.go
+++ b/models/groups_test.go
@@ -1,38 +1,50 @@
-package models
+package models_test
 
 import (
+	"errors"
 	"fmt"
 	"testing"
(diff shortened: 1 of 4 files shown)


models/groups_test.go
```diff
@@ -1,38 +1,50 @@
-package models
+package models_test
```
rowanseymour, line 1: gradually converting these tests to blackbox so that `testsuite` isn't imported by `models` and can thus import `models`

nicpottier (approved):

PR #364, opened 2020-10-01
🔬 Add contact/resolve endpoint to assist with channel events still handled in RP

rowanseymour:


testsuite/db.go
```diff
@@ -3,27 +3,37 @@ package testsuite
...
 )
 
+// MockDB is a mockable proxy to a real sqlx.DB that implements models.Queryer
 type MockDB struct {
-	real       *sqlx.DB
-	callCounts map[string]int
-	shouldErr  func(funcName string, call int) error
+	real            *sqlx.DB
+	callCounts      map[string]int
+	callCountsMutex *sync.Mutex
```
rowanseymour, line 15: unrelated - just noticed this class can panic if we end up with two threads writing to `callCounts` at exactly the same time

web/contact/contact.go
```diff
@@ -128,12 +130,6 @@ func handleModify(ctx context.Context, s *web.Server, r *http.Request) (interfac
 		return nil, http.StatusInternalServerError, errors.Wrapf(err, "unable to load org assets")
 	}
 
-	// clone it as we will modify flows
```
rowanseymour, line 131: unrelated - this appears to be a turd from copy and pasting from the simulator endpoint

nicpottier (approved):
Why is channel required?

rowanseymour:
@nicpottier to set affinity on the URN

nicpottier:
But is that always required? Say adding a RapidPro contact manually before a channel exists. Take it that's a different api path?

rowanseymour:
Yeah this endpoint is specifically for a channel interaction - and to be clear it's only for the relayer code in RP that will move to mailroom

PR #368, opened 2020-10-08
🎏 Enable retrying on the elastic client

rowanseymour:
Very occasionally a call to elastic fails

### DIFF
--- a/mailroom.go
+++ b/mailroom.go
@@ -185,10 +185,7 @@ func (mr *Mailroom) Start() error {
 	}
 
 	// initialize our elastic client
-	mr.ElasticClient, err = elastic.NewClient(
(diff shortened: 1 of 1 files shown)


mailroom.go
```diff
@@ -236,3 +233,16 @@ func (mr *Mailroom) Stop() error {
 	logrus.Info("mailroom stopped")
 	return nil
 }
+
+func newElasticClient(url string) (*elastic.Client, error) {
+	// enable retrying
+	backoff := elastic.NewSimpleBackoff(500, 1000, 2000)
```
rowanseymour, line 239: don't want to tie up RP web threads for too long if there is a problem.. this feels safe 🤷

nicpottier (approved):

PR #376, opened 2020-10-23
Make sure we save attachment file extension

norkans7:


rowanseymour:
@norkans7 so this is fixing the IVR attachments with broken links issue? What exactly is happening there? Is Twilio giving us media files without extensions or without content-type in the HTTP response?

norkans7:
Yes, the media URLs are without extension.

rowanseymour:
From https://www.twilio.com/docs/voice/twiml/record

 > A request to the RecordingUrl will return a recording in binary WAV audio format by default. To request the recording in MP3 format, append ".mp3" to the RecordingUrl.

Wouldn't it be easier to just add .mp3 to URL, then save an attachment with `audio/mp3` content type?

norkans7:
I think we can do that in the IVR case probably, this fix is the same approach we use on courier when do not know the extension, adding to always request mp3 for IVR for twilio is fine too.

rowanseymour:
Mailroom only handles attachments from Twilio, Nexmo and the ticketing services.. so if they give us explicit content-types those seem preferable to sniffing. If we find a case where that doesn't work we can add sniffing.

core/ivr/twiml/twiml.go
```diff
@@ -251,7 +251,7 @@ func (c *client) InputForRequest(r *http.Request) (string, utils.Attachment, err
 		if url == "" {
 			return "", ivr.NilAttachment, nil
 		}
-		return "", utils.Attachment("audio:" + r.Form.Get("RecordingUrl")), nil
+		return "", utils.Attachment("audio:" + url + ".mp3"), nil
```
rowanseymour: audio/mp3:

rowanseymour (changes requested):

nicpottier:
So was this a change on the Twilio side?

norkans7:
I think it has always been so

nicpottier:
So what caused the new issue from Tim noticed?

norkans7:
They use Zapier and Zapier probably expect extensions always, I think that is how they noticed
On our web one can download and listen the WAV file, but not preview it in the browser

rowanseymour (approved):

PR #377, opened 2020-10-23
🏷️ Fix input_labels_added event handling when session input ID isn't set

### LINKED ISSUE #373: Cannot add input label during the expiration/timeout task
We're facing the following error when the resume of a flow try to add some label to a contact input:

> error resuming flow for timeout: error updating session for resume: error applying events: <session_id>: cannot add label, no incoming message for scene: <session_id>

What I found:

1. `"cannot add label, no incoming message for scene: <session_id>"` comes from [here](https://github.com/nyaruka/mailroom/blob/v5.7.33/core/handlers/input_labels_added.go#L41);
2. The `Session.incomingMsgID` is only set [here](https://github.com/nyaruka/mailroom/blob/master/core/tasks/handler/worker.go#L619), when the contact sends a message;
3. Since we don't have any contact input in the expiration task, the `Session.incomingMsgID` isn't set, so if you have a `Add Labels` card after the point of expiration, the resume will fail, leading to an accumulation of flow runs to expire.

## How to reproduce

1. One way is having two flows: a first flow with a `Enter another flow` card calling a second flow that will expire, and a `Add Labels` card trying to label the last contact input:

image

image

2. Set the flow B expiration to minimum (5 minutes):

image

3. Run:

image

4. My log:

image

rowanseymour:
Fixes https://github.com/nyaruka/mailroom/issues/373

### DIFF
--- a/core/handlers/input_labels_added.go
+++ b/core/handlers/input_labels_added.go
@@ -20,12 +20,27 @@ func init() {
 
 // handleInputLabelsAdded is called for each input labels added event in a scene
 func handleInputLabelsAdded(ctx context.Context, tx *sqlx.Tx, rp *redis.Pool, oa *models.OrgAssets, scene *models.Scene, e flows.Event) error {
+	if scene.Session() == nil {
+		return errors.Errorf("cannot add label, not in a session")
+	}
+
 	event := e.(*events.InputLabelsAddedEvent)
 	logrus.WithFields(logrus.Fields{
 		"contact_uuid": scene.ContactUUID(),
 		"session_id":   scene.SessionID(),
 		"labels":       event.Labels,
 	}).Debug("input labels added")
+
+	// in the case this session was started/resumed from a msg event, we have the msg ID cached on the session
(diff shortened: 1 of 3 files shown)


core/handlers/input_labels_added.go
```diff
@@ -20,32 +20,36 @@ func init() {
...
+
 	event := e.(*events.InputLabelsAddedEvent)
 	logrus.WithFields(logrus.Fields{
 		"contact_uuid": scene.ContactUUID(),
 		"session_id":   scene.SessionID(),
 		"labels":       event.Labels,
 	}).Debug("input labels added")
 
+	// in the case this session was started/resumed from a msg event, we have the msg ID cached on the session
+	inputMsgID := scene.Session().IncomingMsgID()
```
rowanseymour, line 35: I had a go at just adding input ID to flow sessions in the database because it seemed that would also solve the responseTo issue, but then realised that actually needs the external ID of the message - us linking messages in the database isn't actually that useful, so then went with this.

nicpottier (approved):

PR #381, opened 2020-11-11
Only update runs with expires on in the future

norkans7:


nicpottier:
Can you give more context for this?

norkans7:
The engine and the expiration task are trying to update the same row at the same time and that will cause the transaction to roll back and the archiver will fail to archive those runs.

https://sentry.io/organizations/nyaruka/issues/1948479022/?project=1285856&query=is%3Aunresolved

https://sentry.io/organizations/nyaruka/issues/1948479054/?project=1285856&query=is%3Aunresolved

https://sentry.io/organizations/nyaruka/issues/586214771/?project=1205568&query=is%3Aunresolved

My understanding here is that we allow the engine to update runs having expires_on in the future so its query and the expirations tasks will always be operating to different rows

nicpottier:
Not updating the run seems way worse. We then have moved the run forward in the session but aren't updating the DB? That's not what we want.

norkans7:
So we fail the session or expire the run in the sprint?

nicpottier:
I'm not actually sure what the right fix is. If this is a race between us resuming a run and expiring a run then we either need some kind of lock or we need to deal with the deadlock properly. IE, if the transaction fails in updating the run here then we need to make sure the session isn't being updated as well.

We need to look at more scenarios to really see what the right thing to do is based on real life examples.

PR #384, opened 2020-11-20
👮 Configure engine to disallow HTTP requests to private networks

rowanseymour:


nicpottier (approved):
Maybe `DisallowedNetworks` would be worth considering?

PR #389, opened 2020-12-11
🤦 Fix misreporting created contacts as updated during imports

rowanseymour:
I think I just forgot to implement this since there's plenty of tests asserting the wrong values


web/contact/testdata/resolve.json
```diff
@@ -49,7 +49,8 @@
             "urn": {
                 "id": 10001,
                 "identity": "tel:+16055742222"
-            }
+            },
+            "created": false
```
rowanseymour, line 53: adding this isn't required.. but seems natural this endpoint would return it

PR #390, opened 2020-12-15
Variable timeout for fire campaign task

nicpottier:
Should help with flow starts that have timing out webhooks


rowanseymour (approved):

nicpottier:
A perhaps more clever way to do this in the future if we see this is to have our runner be aware of how much time it has left when starting a group of starts. It could then bail / commit when say it only has a minute left. This would let us widdle town a set of starts that is time consuming instead of the current state which is a bit all or nothing.

PR #393, opened 2021-01-06
🤫 Add support for passive flows

rowanseymour:
Not much to do here as we already have the concept of not interrupting sessions from campaign single-message flows

### DIFF
--- a/core/models/flows.go
+++ b/core/models/flows.go
@@ -33,6 +33,11 @@ const (
 	FlowTypeVoice     = FlowType("V")
 )
 
+// Interrupts returns whether this flow type interrupts existing sessions
+func (t FlowType) Interrupts() bool {
+	return t != FlowTypePassive
+}
+
 const (
 	flowConfigIVRRetryMinutes = "ivr_retry"
 )
--- a/core/models/flows.go
+++ b/core/models/flows.go
@@ -27,9 +27,10 @@ type FlowType string
 
 // flow type constants
 const (
-	IVRFlow       = FlowType("V")
-	MessagingFlow = FlowType("M")
-	SurveyorFlow  = FlowType("S")
+	FlowTypeMessaging = FlowType("M")
+	FlowTypePassive   = FlowType("P")
+	FlowTypeSurveyor  = FlowType("S")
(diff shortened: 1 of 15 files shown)


core/runner/runner.go
```diff
@@ -226,7 +226,7 @@ func StartFlowBatch(
 	options := NewStartOptions()
 	options.RestartParticipants = batch.RestartParticipants()
 	options.IncludeActive = batch.IncludeActive()
-	options.Interrupt = true
+	options.Interrupt = flow.FlowType() != models.FlowTypePassive
```
rowanseymour: So could still surface exclude_active/include_active in the UI for passive flows.. the flow type only determines whether we interrupt or not if you do include active contacts. Not completely sure that's useful but this leaves the door open.

core/models/flows.go
```diff
@@ -27,19 +27,21 @@ type FlowType string
...
 const (
 	flowConfigIVRRetryMinutes = "ivr_retry"
 )
 
 var flowTypeMapping = map[flows.FlowType]FlowType{
-	flows.FlowTypeMessaging:        MessagingFlow,
-	flows.FlowTypeVoice:            IVRFlow,
-	flows.FlowTypeMessagingOffline: SurveyorFlow,
+	flows.FlowTypeMessaging:        FlowTypeMessaging,
+	flows.FlowTypeMessagingPassive: FlowTypePassive,
```
nicpottier, line 47: Minor but wonder if `PassiveMessaging` might be a bit better? It is still messaging after all. (only say this because the type does still imply the medium, which in this case is still messaging)

core/runner/runner.go
nicpottier: Alternate random idea: `FlowType` has an `Interrupt()` on it so this just becomes `flow.FlowType().Interrupt()`

Just a thought though.

nicpottier (approved):

core/models/flows.go
rowanseymour: then I start thinking why don't they match the engine types.. why `FlowTypeSurveyor` over `FlowTypeMessagingOffline`.. so I convinced myself to just let these match the database codes M, V, P, S

PR #395, opened 2021-01-12
📡 Rework message sending, fix not triggering FCM sync for non-flow messages

rowanseymour:


core/msgio/android.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+var clientInit sync.Once
+var fcmClient *fcm.Client
+
+func init() {
+	clientInit.Do(func() {
+		if config.Mailroom.FCMKey == "" {
+			logrus.Error("fcm not configured, no syncing of android channels")
```
rowanseymour, line 21: so now you see this on mailroom startup, and after that you only see warnings when syncing Android messages is skipped

core/msgio/android.go
```diff
@@ -0,0 +1,69 @@
...
+			return
+		}
+
+		var err error
+		fcmClient, err = fcm.NewClient(config.Mailroom.FCMKey)
+		if err != nil {
+			panic(errors.Wrap(err, "unable to create FCM client"))
```
rowanseymour, line 28: this can't actually happen with the current version of that library but who knows about the future

core/msgio/courier.go
```diff
@@ -1,13 +1,14 @@
-package courier
+package msgio
```
rowanseymour, line 1: Needed a package name for both the courier queuing and android syncing.. technically the `i` is a lie but couldn't think of anything better. Tried `msgs` but thats a very common variable name.

core/msgio/android.go
rowanseymour: alternatively this client could be added to the `Mailroom` instance like we do for other clienty things.. that's arguably a cleaner way to initialize something on startup that depends on the config..

rowanseymour: @nicpottier I think that's the actually the best approach and it makes mocking in tests super simple.. but it means passing FCM client instances around everywhere so for the sake of this diff maybe I'll address that in a subsequent PR if you don't think it's a terrible idea

core/msgio/send.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// walk through our messages, separate by whether they have a channel and if it's Android
+	for _, msg := range msgs {
+		channel := msg.Channel()
+		if channel != nil {
+			if channel.Type() == models.ChannelTypeAndroid {
+				if !androidChannelsSeen[channel] {
+					androidChannels = append(androidChannels, channel)
+				}
+				androidChannelsSeen[channel] = true
```
nicpottier, line 32: Minor style but feels clearer to me at least to only have the "set" here and then convert to the array if it has values before calling Sync.

core/msgio/send.go
```diff
@@ -0,0 +1,79 @@
...
+			} else {
+				courierMsgs[msg.ContactID()] = append(courierMsgs[msg.ContactID()], msg)
+			}
+		} else {
+			pending = append(pending, msg)
+		}
+	}
+
+	fmt.Printf("%v\n", androidChannels)
+	fmt.Printf("%v\n", androidChannelsSeen)
```
nicpottier: crud

nicpottier (approved):

core/msgio/send.go
rowanseymour: I think I prefer this because it means one loop and it preserves order - which is not a huge deal but makes the test assertions easier. Tho if go ever gave us a generic `keys(..)` method I'd feel differently..

PR #396, opened 2021-01-13
🧯 Fix updating message status when id column is bigint

rowanseymour:
I don't think mailroom commonly updates message status thus why we haven't seen this in sentry. It's only when queuing flow messages to courier fails, or receiving channel-less messages, that we put the failed messages into status PENDING.

### DIFF
--- a/core/models/msgs.go
+++ b/core/models/msgs.go
@@ -23,7 +23,6 @@ import (
 	"github.com/nyaruka/null"
 
 	"github.com/gomodule/redigo/redis"
-	"github.com/jmoiron/sqlx"
 	"github.com/lib/pq"
 	"github.com/lib/pq/hstore"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


core/models/msgs.go
```diff
@@ -460,25 +459,19 @@ func UpdateMessage(ctx context.Context, tx Queryer, msgID flows.MsgID, status Ms
...
 
 // MarkMessagesPending marks the passed in messages as pending
-func MarkMessagesPending(ctx context.Context, tx *sqlx.Tx, msgs []*Msg) error {
-	return updateMessageStatus(ctx, tx, msgs, MsgStatusPending)
+func MarkMessagesPending(ctx context.Context, db Queryer, msgs []*Msg) error {
+	return updateMessageStatus(ctx, db, msgs, MsgStatusPending)
 }
 
-// MarkMessagesQueued marks the passed in messages as queued
-func MarkMessagesQueued(ctx context.Context, tx *sqlx.Tx, msgs []*Msg) error {
```
rowanseymour, line 468: not used anywhere

core/models/msgs.go
```diff
@@ -491,7 +484,7 @@ FROM (
 ) AS
 	m(id, status)
 WHERE
-	msgs_msg.id = m.id::int
+	msgs_msg.id = m.id::bigint
```
rowanseymour, line 487: without this change, the second `MarkMessagesPending` call in the unit test errors with `pq: value "3000000000" is out of range for type integer`

core/models/msgs_test.go
```diff
@@ -166,3 +166,47 @@ func TestNormalizeAttachment(t *testing.T) {
...
+	msg2 := insertMsg("Hola")
+	insertMsg("Howdy")
+
+	models.MarkMessagesPending(ctx, db, []*models.Msg{msg1, msg2})
+
+	testsuite.AssertQueryCount(t, db, `SELECT count(*) FROM msgs_msg WHERE status = 'P'`, nil, 2)
+
+	// try running on database with BIGINT message ids
+	db.MustExec(`ALTER TABLE "msgs_msg" ALTER COLUMN "id" TYPE bigint USING "id"::bigint;`)
+	db.MustExec(`ALTER SEQUENCE "msgs_msg_id_seq" AS bigint;`)
```
rowanseymour, line 201: not sure why when we did this on prod, the sequences were already bigints. That is the default in Postgres but not in the mailroom test db. Without this line, next ALTER errors with:

```
pq: RESTART value (3000000000) cannot be greater than MAXVALUE (2147483647) 
```

core/models/msgs.go
rowanseymour: I was concerned that if that database column is `int` then this might cause it to not use an index built on that column, but ran some tests with queries using `::bigint` and it doesn't make a difference.

nicpottier (approved):

PR #397, opened 2021-01-14
FCM client refactor

rowanseymour:
Moves to `Mailroom` instance alongside the elastic client, redis pool etc instead of being a global.

### DIFF
--- a/core/models/events.go
+++ b/core/models/events.go
@@ -238,7 +238,7 @@ func HandleAndCommitEvents(ctx context.Context, db QueryerWithTx, rp *redis.Pool
 }
 
 // ApplyModifiers modifies contacts by applying modifiers and handling the resultant events
-func ApplyModifiers(ctx context.Context, db QueryerWithTx, rp *redis.Pool, fc *fcm.Client, oa *OrgAssets, modifiersByContact map[*flows.Contact][]flows.Modifier) (map[*flows.Contact][]flows.Event, error) {
+func ApplyModifiers(ctx context.Context, db QueryerWithTx, oa *OrgAssets, modifiersByContact map[*flows.Contact][]flows.Modifier) (map[*flows.Contact][]flows.Event, error) {
 	// create an environment instance with location support
 	env := flows.NewEnvironment(oa.Env(), oa.SessionAssets().Locations())
 
@@ -253,7 +253,7 @@ func ApplyModifiers(ctx context.Context, db QueryerWithTx, rp *redis.Pool, fc *f
 		eventsByContact[contact] = events
 	}
 
-	err := HandleAndCommitEvents(ctx, db, rp, fc, oa, eventsByContact)
+	err := HandleAndCommitEvents(ctx, db, nil, nil, oa, eventsByContact)
 	if err != nil {
 		return nil, errors.Wrap(err, "error commiting events")
 	}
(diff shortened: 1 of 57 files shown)


core/models/imports.go
```diff
@@ -113,7 +113,7 @@ func (b *ContactImportBatch) tryImport(ctx context.Context, db *sqlx.DB, orgID O
 	}
 
 	// and apply in bulk
-	_, err = ApplyModifiers(ctx, db, nil, oa, modifiersByContact)
+	_, err = ApplyModifiers(ctx, db, nil, nil, oa, modifiersByContact)
```
rowanseymour, line 116: imports don't send messages

web/surveyor/surveyor.go
```diff
@@ -133,7 +133,7 @@ func handleSubmit(ctx context.Context, s *web.Server, r *http.Request) (interfac
 	if err != nil {
 		return nil, http.StatusInternalServerError, errors.Wrapf(err, "error starting transaction for session write")
 	}
-	sessions, err := models.WriteSessions(ctx, tx, s.RP, oa, []flows.Session{fs}, []flows.Sprint{sprint}, nil)
+	sessions, err := models.WriteSessions(ctx, tx, s.RP, nil, oa, []flows.Session{fs}, []flows.Sprint{sprint}, nil)
```
rowanseymour, line 136: surveyor sessions shouldn't be sending android messages

core/models/events.go
```diff
@@ -237,7 +238,7 @@ func HandleAndCommitEvents(ctx context.Context, db QueryerWithTx, rp *redis.Pool
 }
 
 // ApplyModifiers modifies contacts by applying modifiers and handling the resultant events
-func ApplyModifiers(ctx context.Context, db QueryerWithTx, rp *redis.Pool, oa *OrgAssets, modifiersByContact map[*flows.Contact][]flows.Modifier) (map[*flows.Contact][]flows.Event, error) {
```
rowanseymour, line 240: modifiers don't send messages so don't need redis (for talking to courier) or the FCM client (for talking to Android)

nicpottier (approved):
I'm not sure this feels better to me, the hook interface changing seems especially ugly now, but if you really want it, go for it.

rowanseymour:
I started to have my doubts realising how many places need to pass it around but I _think_ I still prefer the functional approach and if we're worried about function argument bloat we could possibly pack things like this into a context structure.

nicpottier:
Does FCMClient really have any state that is interesting? Maybe our problem is having that be an object in the first place?

nicpottier:
Looking at it it is basically just an http.Client, so it really isn't maintaining much state over just the config. I don't know, it isn't the end of the world for sure, but just really feels wrong to have the Hook stuff even know about FCMClient, especially given how rarely it is used.

rowanseymour:
What's the difference between hooks knowing about redis.Pool and knowing about fcm.Client ?

nicpottier:
To me it is that redis/sql are more generically used by other hooks. They also actually have state beyond just configs (pools for the connections).

rowanseymour:
Ok so what do you want to do now that this is merged? I don't feel very strongly in favor of this so if you feel it's really wrong then I can revert it.

nicpottier:
I approved it so I can live with it. :)

rowanseymour:
I've reverted it for now so can ponder on it a bit more

PR #398, opened 2021-01-14
FCM client improvements take II

rowanseymour:


core/msgio/android.go
```diff
@@ -12,27 +11,9 @@ import (
...
 )
 
-var clientInit sync.Once
-var fcmClient *fcm.Client
-
-func init() {
-	clientInit.Do(func() {
-		if config.Mailroom.FCMKey == "" {
-			logrus.Error("fcm not configured, no syncing of android channels")
-			return
```
rowanseymour, line 22: this moves into mailroom instance startup - but just the warning - so it's not the first thing to appear in the console

nicpottier (approved):
yay!

PR #400, opened 2021-01-20
Elastic v7

nicpottier:


core/models/search.go
```diff
@@ -77,7 +77,7 @@ func ContactIDsForQueryPage(ctx context.Context, client *elastic.Client, org *Or
 		return nil, nil, 0, errors.Wrapf(err, "error parsing sort")
 	}
 
-	s := client.Search("contacts").Routing(strconv.FormatInt(int64(org.OrgID()), 10))
+	s := client.Search("contacts").TrackTotalHits(true).Routing(strconv.FormatInt(int64(org.OrgID()), 10))
```
nicpottier, line 80: Without this we may get "estimates" on the total # of hits. This forces true counts.

rowanseymour: Interesting - I wonder what the performance differences are. Could be useful in scenarios like the flow start modal where we could give users an estimate.

rowanseymour (approved):

core/models/search.go
nicpottier: It doesn't really give an estimate, rather it'll return something like "> 10,000", ie, stop counting at a particular point. But ya, for the case of gathering the union of contacts in groups elastic is probably pretty great. The real challenge with that dialog is which contacts are active in what flows and which have already participated etc..

PR #401, opened 2021-01-25
🐘 Test on PG 11 and 12

rowanseymour:


.github/workflows/ci.yml
```diff
@@ -19,26 +20,32 @@ jobs:
...
     - name: Install Linux packages
       run: sudo apt install -y --no-install-recommends postgresql-client
 
     - name: Initialize database
+      # we create our test database with a different user so that we can drop everything owned by this user between tests
       run: |
-        psql -h localhost -U postgres -c "CREATE USER mailroom_test PASSWORD 'temba';"
-        psql -h localhost -U postgres -c "ALTER ROLE mailroom_test WITH SUPERUSER;"
-        psql -h localhost -U postgres -c "CREATE DATABASE mailroom_test;"
+        export PGPASSWORD=temba
```
rowanseymour, line 35: The docker image with postgres is now postgis:postgis instead of mdillon:postgis and it seems this new one doesn't set POSTGRES_HOST_AUTH_METHOD=trust so you need to set a password for the superuser account and then set it here as an env for psql to use.

core/models/channel_connection.go
```diff
@@ -251,6 +251,9 @@ WHERE
 	cc.channel_id = $1 AND
 	cc.connection_type = $2 AND
 	cc.external_id = $3
+ORDER BY
+	cc.id DESC
+LIMIT 1
```
rowanseymour, line 256: @nicpottier `TestTwilioIVR` was failing for PG 12 but after much debugging and thinking oh crap we've written some incompatible SQL somewhere it seems it's a case of `SelectChannelConnectionByExternalID` "always" returning the last of multiple matching connections when tests run under PG 10 and 11, but the first for PG 12. I don't understand the IVR stuff well enough yet to know if it should be matching multiple connections, but this at least makes the assumed behavior explicit.

core/models/orgs.go
```diff
@@ -264,7 +264,7 @@ SELECT ROW_TO_JSON(o) FROM (SELECT
 	(SELECT CASE is_anon WHEN TRUE THEN 'urns' WHEN FALSE THEN 'none' END) AS redaction_policy,
 	$2::int AS max_value_length,
 	(SELECT iso_code FROM orgs_language WHERE id = o.primary_language_id) AS default_language,
-	(SELECT ARRAY_AGG(iso_code) FROM orgs_language WHERE org_id = o.id) AS allowed_languages,
+	(SELECT ARRAY_AGG(iso_code ORDER BY iso_code ASC) FROM orgs_language WHERE org_id = o.id) AS allowed_languages,
```
rowanseymour, line 267: also was causing failing test due to non-deterministic language ordering

PR #406, opened 2021-02-09
Should match referral trigger with case insensitive

norkans7:


norkans7:
replaced by https://github.com/nyaruka/mailroom/pull/407

PR #409, opened 2021-02-17
Rename tickets/internal package

### LINKED ISSUE #408: Internal ticket type not found, rename?
The internal ticket type was not found when executing flows. After a bunch of poking about (being the newb that I am), I'm guessing it just needs to be added to the imports in `cmd/go/main.go`. However, doing so immediately exits since it looks like `internal` packages have special meaning in golang.

https://golang.org/doc/go1.4#internalpackages

rowanseymour:
Don't really want to rename our concept of an internal ticketer just because of a golang quirk.. so just renaming the package a little to fix #408


nicpottier (approved):
Get that intern on it!

PR #411, opened 2021-02-17
Minor copypasta

ericnewcomer:
Just a couple copy paste bits on some tests I saw while I was poking around

### DIFF
--- a/web/ticket/ticket_test.go
+++ b/web/ticket/ticket_test.go
@@ -17,8 +17,8 @@ func TestTicketClose(t *testing.T) {
 	testsuite.Reset()
 	db := testsuite.DB()
 
-	// create 2 open tickets and 1 closed one for Cathy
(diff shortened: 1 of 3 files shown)


web/ticket/ticket_test.go
```diff
@@ -18,7 +18,7 @@ func TestTicketClose(t *testing.T) {
 	db := testsuite.DB()
 
 	// create 2 open tickets and 1 closed one for Cathy
-	testdata.InsertOpenTicket(t, db, models.Org1, models.CathyID, models.MailgunID, flows.TicketUUID(uuids.New()), "Need help", "Have you seen my cookies?", "17")
+	testdata.InsertOpenTicket(t, db, models.Org1, models.CathyID, models.ZendeskID, flows.TicketUUID(uuids.New()), "Need help", "Have you seen my cookies?", "17")
```
rowanseymour: I think it's intentional here that we test with tickets from different ticketers

ericnewcomer: ya, i think i goofed that one, im double checking

PR #413, opened 2021-02-22
IVR Forwards

nicpottier:


core/models/runs.go
```diff
@@ -861,6 +861,11 @@ WHERE
 
 // FindActiveSessionOverlap returns the list of contact ids which overlap with those passed in which are active in any other flows
 func FindActiveSessionOverlap(ctx context.Context, db *sqlx.DB, flowType FlowType, contacts []ContactID) ([]ContactID, error) {
+	// background flows should look at messaging flows when determing overlap (background flows can't be active by definition)
+	if flowType == FlowTypeBackground {
+		flowType = FlowTypeMessaging
```
nicpottier, line 866: This is the fix for background flows not honoring skipping active flows when being started. (by say a campaign)

core/ivr/ivr.go
```diff
@@ -424,126 +426,159 @@ func ResumeIVRFlow(
 	}
 
 	if body != nil {
+		// guess our content type and set it
+		contentType := http.DetectContentType(body)
```
rowanseymour, line 430: We use this elsewhere but if I recall right it's not very smart and basically returns `text/html`, `text/plain` or `application/octet-stream`. Not sure if that's important here.

rowanseymour (approved):

PR #416, opened 2021-03-10
Fail expirations that are no longer the active session

nicpottier:
This update has us fail the runs and sessions that are expired when there is a newer active session. Technically this shouldn't occur, so we log an error, but we may as well fail the old run and session while we are at it and perhaps we can start seeing patterns when this occurs to track down the possible race causing these to occur. (we have a few dozen in total on TextIt)

Note this does introduce us using the FAILED status on session and run, which we previously weren't using. I feel this is a more accurate status than say "interrupted", which we have used previous for cases like this. Another similar case is a flow resuming for a flow that has been deleted. Previously we were updating the run and session to interrupted in those cases, but FAILED seems more right to me? Totally open to discussion there.


rowanseymour (approved):
Looks good, just need https://github.com/nyaruka/rapidpro/pull/3224 first

PR #418, opened 2021-03-18
Smarter loading for cache misses on org assets

nicpottier:
So a problem we have on the mailroom side is our thundering herd on cache misses when loading org assets. We evict cached org assets every five seconds and that can cause issues when we have a high concurrency of events on a single org, such as when an org is very busy or even when we self impose this with per-minute timeouts and expirations.

Basically the problem is if we have 12 goroutines running through all the expirations for an org, they are all going to get a cache miss at the same time and try to load new assets all at once. This only gets worse the more load the DB has since it starts extending how long a full asset load takes and therefore encourages more misses.

Solving this in an idiomatic golang way is actually a bit tricky. Explored a bunch of different options but in the end found the same problem being solved in LakeFS (https://github.com/treeverse/lakeFS/blob/10c7fcf914d423d4b590ed1f9773b7ef68f4c092/pkg/cache/only_one.go) Just adapted that for our specific use case since we don't really need a generic solution here.

Concept is basically:
 * use a sync.Map to create/get only one "loader" per org as needed
 * goroutines use the channel on that "loader" to block on to wait for the load to complete
 * "loader" also stores result/error so they can access that once channel is closed

This should keep us from having concurrent / herd problems in loading the assets.

### DIFF
--- a/core/models/assets.go
+++ b/core/models/assets.go
@@ -74,7 +74,7 @@ var ErrNotFound = errors.New("not found")
 // we cache org objects for 5 seconds, cleanup every minute (gets never return expired items)
 var orgCache = cache.New(time.Second*5, time.Minute)
 
-// map of org id -> chanAndResult used to make sure we are only loading one org at a time
+// map of org id -> assetLoader used to make sure we only load an individual org once when expired
 var assetLoaders = sync.Map{}
 
 // represents a goroutine loading assets for an org, stores the loaded assets (and possible error) and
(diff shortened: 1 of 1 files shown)


norkans7 (approved):

core/models/assets.go
```diff
@@ -74,6 +74,33 @@ var ErrNotFound = errors.New("not found")
 // we cache org objects for 5 seconds, cleanup every minute (gets never return expired items)
 var orgCache = cache.New(time.Second*5, time.Minute)
 
+// map of org id -> chanAndResult used to make sure we are only loading one org at a time
```
rowanseymour: comment is slightly misleading.. it's.. an org.. once

rowanseymour (approved):

Issue #177, opened 2019-09-18
Task to send non-broadcast messages


rowanseymour:
There are two places where RP is still queueing messages in courier:

 1. User clicks on a failed message and retries it - this creates a copy of the message and queues that for sending in courier
 2. User adds a topup and that is applied to any unsent messages, and those are then queued to courier.

If we were to move that into mailroom, it could be task that mirrors `Org.trigger_send` in RP:

```
{
    "type": "queue_msgs"
    "org_id": [12],
    "msgs": [123, 345]
}
```

If `msgs` isn't provided, then all `PENDING` messages with topups are queued. We would tho have to get an org-level lock to ensure we don't double queue messages to courier, but it seems like most of the code is already there in mailroom, i.e. `QueueMessages`

rowanseymour:
Or maybe this task should be the one managing credits? It's not ideal that RP still has to know about tracking and decrementing credits.. just for the case of resending a single failed message.

rowanseymour:
In the case of the message resend - we also re-select the channel which is a fair amount of logic. If that could be done in via a task then would allow us to maintain that logic in only goflow/mailroom. Maybe these are two different mailroom tasks...

rowanseymour:
Resending is now done entirely in mailroom but we still have sending logic in rapidpro for when new topups are added

rowanseymour:
We got rid of that sending code in RP and now only mailroom queues to courier

PR #420, opened 2021-04-23
🔁 Add task to resend messages

rowanseymour:
Part of #177 which would let us remove last of the topup and message sending code in RP

### DIFF
--- a/core/models/msgs.go
+++ b/core/models/msgs.go
@@ -411,13 +411,14 @@ FROM
 	msgs_msg
 WHERE
 	org_id = $1 AND
-	id = ANY($2)
+	direction = $2 AND
+	id = ANY($3)
 ORDER BY
 	id ASC`
 
 // LoadMessages loads the given messages for the passed in org
-func LoadMessages(ctx context.Context, db Queryer, orgID OrgID, msgIDs []MsgID) ([]*Msg, error) {
-	rows, err := db.QueryxContext(ctx, loadMessagesSQL, orgID, pq.Array(msgIDs))
+func LoadMessages(ctx context.Context, db Queryer, orgID OrgID, direction MsgDirection, msgIDs []MsgID) ([]*Msg, error) {
+	rows, err := db.QueryxContext(ctx, loadMessagesSQL, orgID, direction, pq.Array(msgIDs))
 	if err != nil {
 		return nil, errors.Wrapf(err, "error querying msgs for org: %d", orgID)
 	}
--- a/core/models/msgs.go
+++ b/core/models/msgs.go
@@ -9,8 +9,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/nyaruka/gocommon/dates"
 	"github.com/nyaruka/gocommon/gsm7"
 	"github.com/nyaruka/gocommon/urns"
+	"github.com/nyaruka/gocommon/uuids"
 	"github.com/nyaruka/goflow/assets"
 	"github.com/nyaruka/goflow/envs"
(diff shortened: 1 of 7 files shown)


core/models/msgs.go
```diff
@@ -458,12 +514,8 @@ func UpdateMessage(ctx context.Context, tx Queryer, msgID flows.MsgID, status Ms
 	return nil
 }
 
-// MarkMessagesPending marks the passed in messages as pending
-func MarkMessagesPending(ctx context.Context, db Queryer, msgs []*Msg) error {
```
rowanseymour: not sure it's worth having a separate function for each status

nicpottier (changes requested):
Why are we cloning the message? This feels like some remnant of the previous implementation that just adds complication. I'd rather we just re-queue it and make sure any kind of double sending cache or whatnot is cleared so that works. Cloning just seems odd especially given how often this occurs.

rowanseymour:
Good question I was just copying what we had in RapidPro.. maybe we did it this way to ensure it uses another topup? Will rework this to just requeue.

nicpottier:
Was thinking about this on the dog walk and I think it may have to do with the relayer, though that sounds like something perhaps worth revisiting.

rowanseymour:
@nicpottier in order for mailroom to clear the double-send protection, it would have to know the specifics of courier's date-based redis sets we use for that... maybe mailroom queues it to courier with a `is_resend` flag set and that skips double send checking in courier?

nicpottier:
Ya, that makes sense, can look at that today.

rowanseymour:
@nicpottier reworked to reset message state rather than clone messages

core/models/msgs.go
```diff
@@ -949,6 +1006,73 @@ func CreateBroadcastMessages(ctx context.Context, db Queryer, rp *redis.Pool, oa
...
+		// allocate a new topup for this message if org uses topups
+		msg.m.TopupID, err = AllocateTopups(ctx, db, rp, oa.Org(), 1)
+		if err != nil {
+			return errors.Wrapf(err, "error allocating topup for message resending")
+		}
+
+		// mark message as being a resend so it will be queued to courier as such
+		msg.m.Status = MsgStatusPending
+		msg.m.QueuedOn = dates.Now()
+		msg.m.SentOn = dates.ZeroDateTime
```
rowanseymour, line 1060: not sure about this.. I think it'll put a zero time in the json queued to courier.. but that doesn't matter?

nicpottier: Does `omit_empty` leave it out if it is a zero value? Feels risky / weird having a 0 value there.

nicpottier (approved):

core/models/msgs.go
rowanseymour: Apparently not but aren't we always queuing `sent_on="0001-01-01T00:00:00Z"` to courier? I need to verify but I can't see how the regular queuing code would be any different.

rowanseymour: Ok now I've checked and we always queue to courier with `sent_on="0001-01-01T00:00:00Z"` and `next_attempt="0001-01-01T00:00:00Z"` because those are non-null in mailroom's model

nicpottier:
Ya, if that's what we were doing then fine. Maybe this is an argument to add `null.Time` to our null package at some point, feels risky / slopping to be writing real dates.

nicpottier:
Just realized when looking at this further that both `queued_on` and `sent_on` are actually non-null in our DB too, which seems like the first thing to tweak / fix!

PR #432, opened 2021-06-02
🔫 Add support for exclusion groups to triggers 

rowanseymour:
Also generalizes how we select the best match from triggers when they have groups, channels etc, making it really easy to add groups to other trigger types... so I added a lot of tests to make sure nothing is broken

### DIFF
--- a/core/models/triggers.go
+++ b/core/models/triggers.go
@@ -91,60 +91,6 @@ func (t *Trigger) Match() *triggers.KeywordMatch {
 	return nil
 }
 
-type triggerMatch struct {
-	trigger        *Trigger
-	typeScore      int // e.g. keyword (2) vs catchall (1)
-	qualifierScore int // e.g. group inclusion (2) or channel match (4)
-}
-
-func (m *triggerMatch) score() int {
-	return m.typeScore*10 + m.qualifierScore
-}
-
-const triggerScoreByChannel = 4
-const triggerScoreByInclusion = 2
-const triggerScoreByExclusion = 1
-
-// matches against the qualifiers (inclusion groups, exclusion groups, channel) on this trigger and returns a score
-func (t *Trigger) MatchQualifiers(channel *Channel, contactGroups map[GroupID]bool) (bool, int) {
-	score := 0
-
-	if channel != nil && t.t.ChannelID != NilChannelID {
-		if t.ChannelID() == channel.ID() {
-			score += triggerScoreByChannel
-		} else {
-			return false, 0
-		}
-	}
-
-	if len(t.t.IncludeGroupIDs) > 0 {
-		inGroup := false
-		// if contact is in any of the groups to include that's a match by inclusion
-		for _, g := range t.t.IncludeGroupIDs {
-			if contactGroups[g] {
-				inGroup = true
-				score += triggerScoreByInclusion
-				break
-			}
-		}
-		if !inGroup {
-			return false, 0
-		}
-	}
-
-	if len(t.t.ExcludeGroupIDs) > 0 {
-		// if contact is in none of the groups to exclude that's a match by exclusion
-		for _, g := range t.t.ExcludeGroupIDs {
-			if contactGroups[g] {
-				return false, 0
-			}
-		}
-		score += triggerScoreByExclusion
-	}
-
-	return true, score
-}
-
 // loadTriggers loads all non-schedule triggers for the passed in org
 func loadTriggers(ctx context.Context, db Queryer, orgID OrgID) ([]*Trigger, error) {
 	start := time.Now()
@@ -171,14 +117,8 @@ func loadTriggers(ctx context.Context, db Queryer, orgID OrgID) ([]*Trigger, err
 	return triggers, nil
 }
 
-// FindMatchingMsgTrigger returns the matching trigger (if any) for the given message text and contact
-func FindMatchingMsgTrigger(org *OrgAssets, contact *flows.Contact, text string) *Trigger {
-	// build a set of the groups this contact is in
-	groupIDs := make(map[GroupID]bool, 10)
-	for _, g := range contact.Groups().All() {
-		groupIDs[g.Asset().(*Group).ID()] = true
-	}
-
+// FindMatchingMsgTrigger finds the best match trigger for an incoming message from the given contact
+func FindMatchingMsgTrigger(oa *OrgAssets, contact *flows.Contact, text string) *Trigger {
 	// determine our message keyword
 	words := utils.TokenizeString(text)
 	keyword := ""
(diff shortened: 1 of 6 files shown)


core/models/triggers.go
```diff
@@ -294,17 +319,21 @@ SELECT ROW_TO_JSON(r) FROM (SELECT
 	t.match_type as match_type,
 	t.channel_id as channel_id,
 	COALESCE(t.referrer_id, '') as referrer_id,
-	ARRAY_REMOVE(ARRAY_AGG(g.contactgroup_id), NULL) as group_ids
+	ARRAY_REMOVE(ARRAY_AGG(DISTINCT ig.contactgroup_id), NULL) as include_group_ids,
+	ARRAY_REMOVE(ARRAY_AGG(DISTINCT eg.contactgroup_id), NULL) as exclude_group_ids
```
rowanseymour, line 303: without `DISTINCT` you get dupes from having both joins

core/models/triggers.go
```diff
@@ -31,14 +32,14 @@ const (
 	MissedCallTriggerType      = TriggerType("M")
 	NewConversationTriggerType = TriggerType("N")
 	ReferralTriggerType        = TriggerType("R")
-	CallTriggerType            = TriggerType("V")
+	IncomingCallTriggerType    = TriggerType("V")
```
rowanseymour, line 35: just matching names with those in RP

core/models/triggers.go
```diff
@@ -113,176 +117,163 @@ func loadTriggers(ctx context.Context, db Queryer, orgID OrgID) ([]*Trigger, err
...
-func FindMatchingReferralTrigger(org *OrgAssets, channel *Channel, referrerID string) *Trigger {
-	var match *Trigger
-	for _, t := range org.Triggers() {
-		if t.TriggerType() == ReferralTriggerType {
-			// matches referrer id? that takes top precedence, return right away
-			if strings.ToLower(referrerID) != "" && strings.EqualFold(referrerID, t.ReferrerID()) && (t.ChannelID() == NilChannelID || t.ChannelID() == channel.ID()) {
-				return t
-			}
+func FindMatchingReferralTrigger(oa *OrgAssets, channel *Channel, referrerID string) *Trigger {
+	candidates := findTriggerCandidates(oa, ReferralTriggerType, nil)
```
rowanseymour: don't actually check the referrer here as it's a thing that needs scored - a match on a trigger with a non-empty referrer id scores more highly than a match with a trigger with an empty referrer id

core/models/triggers.go
```diff
@@ -113,176 +117,163 @@ func loadTriggers(ctx context.Context, db Queryer, orgID OrgID) ([]*Trigger, err
...
-		groupIDs[g.Asset().(*Group).ID()] = true
+type triggerMatch struct {
+	trigger *Trigger
+	score   int
+}
+
+const triggerScoreByReferrer = 8
+const triggerScoreByChannel = 4
+const triggerScoreByInclusion = 2
+const triggerScoreByExclusion = 1
```
rowanseymour, line 231: this is just a way of being able to sort triggers by a set of criteria

rowanseymour: we could pretty easily now let all triggers have groups etc

norkans7 (approved):

core/models/triggers.go
```diff
@@ -113,176 +117,166 @@ func loadTriggers(ctx context.Context, db Queryer, orgID OrgID) ([]*Trigger, err
...
-	// have a normal match? return that
-	if match != nil {
-		return match
+	if len(t.IncludeGroupIDs()) > 0 {
+		inGroup := false
+		// if contact is in any of the groups to include that's a match by inclusion
+		for _, g := range t.IncludeGroupIDs() {
+			if contactGroups[g] {
+				inGroup = true
+				score += triggerScoreByInclusion
```
nicpottier, line 271: huh, so that means that given two triggers one on one group and one on two groups if the contact is in both groups then it'll match the second. Which makes as much sense as anything I suppose?

nicpottier (approved):

core/models/triggers.go
```diff
@@ -294,17 +288,21 @@ SELECT ROW_TO_JSON(r) FROM (SELECT
...
+	LEFT OUTER JOIN triggers_trigger_groups ig ON t.id = ig.trigger_id
+	LEFT OUTER JOIN triggers_trigger_exclude_groups eg ON t.id = eg.trigger_id
 WHERE 
 	t.org_id = $1 AND 
 	t.is_active = TRUE AND
 	t.is_archived = FALSE AND
 	t.trigger_type != 'S'
 GROUP BY 
 	t.id
+ORDER BY
```
nicpottier: Is there any argument to include keyword here before id?

nicpottier (approved):

core/models/triggers.go
rowanseymour: ahh no.. not my intention.. the ordering should be

```
channel (4) + include (2) + exclude (1) = 7
channel (4) + include (2) = 6
channel (4) + exclude (1) = 5
channel (4) = 4
include (2) + exclude (1) = 3
include (2) = 2
exclude (1) = 1
```

which I'll add as a comment in the code so future generations can understand what is happening!

core/models/triggers.go
rowanseymour: ordering here isn't actually deterministic since we find matching triggers and sort them in go without looking at ID or anything else.. which is maybe why we need an explicit fallback ordering like created_on or a new priority field

core/models/triggers.go
nicpottier: Ah, missed the break, but missing a break in exclude then or *that* one works that way? I'm not entirely sure exclusion should participate in scoring though, that feels more like a filter rather than a score attribute. IE, two keywords on different groups that both match but one has an exclusion I'm not sure that really makes sense that the exclusion one ranks higher? (as opposed to lexical?)

rowanseymour: Really? I think for two otherwise identical triggers, it would be natural to prioritize the one with an exclude, e.g. 

1. keyword="help" include=U-Reporters exclude=Open Tickets
2. keyword="help" include=U-Reporters

rowanseymour: Tho I have just noticed that we don't allow include groups to be repeated across triggers with the same keyword (I thought it was that the set of the groups couldn't be the same).. so now I'm not sure if we'd ever allow the example of (U-Reporters being repeated). 

So I guess the question is are we going to allow something like this:

```
1. keyword="help" exclude=Open Tickets
2. keyword="help"
```

like how we allow the following and the one with the group takes precedence:

```
1. keyword="help" include=Open Tickets
2. keyword="help"
```

PR #434, opened 2021-06-03
write sessions to s3 on resumes (optionally)

nicpottier:
Right now just for resumes.. some open questions/todos:
 * Do we want to write sessions to the same bucket as attachments, etc.. `dl-rapidpro-io` currently. They are both private by default though `dl-rapidpro-io` CAN have public access so we could further lock down sessions. Kind of feels like it should be a separate bucket.. but..
 * Passing the storage down to everything is one more thing to pass down as an argument. Feels like maybe we should just pass down an interface that lets you get the db, redis pool, media storage, session storage etc.. this would be easy except we'd have to come up with a name for it which is impossible! :) `MailroomContext` passed as `mrc`?
  * Adding this to new sessions seems pretty straightfoward just thought starting with resumes would be lowest risk 
  
  @rowanseymour thoughts?


nicpottier:
@rowanseymour think this is good to go now (need to merge https://github.com/nyaruka/rapidpro/pull/3336 first though)

core/models/orgs.go
```diff
@@ -90,6 +92,11 @@ func (o *Org) Suspended() bool { return o.o.Suspended }
 // UsesTopups returns whether the org uses topups
 func (o *Org) UsesTopups() bool { return o.o.UsesTopups }
 
+func (o *Org) UsesStorageSessions() bool {
+	// true is true, everything else is false
+	return o.ConfigValue(configStorageSessions, "false") == "true"
```
rowanseymour: do we want to make this something like "w" / "rw" so we can reuse for step 2?

rowanseymour (approved):

core/models/orgs.go
nicpottier: good idea.. tweaked it to use `session_storage_mode` as the config value with `db`, `s3` and `s3_write` as the possible values.

PR #435, opened 2021-06-04
introduce runtime, move Server to use it

nicpottier:
Added use in Mailroom and tasks as well. I think from here on out we can switch things over as needed.

Maybe rule is if you are passing more than just two items in the Runtime then pass Runtime instead. (IE, if you only care about db, pass that, or even if you use db and rp, pass those in, but if more than that pass in runtime)

### DIFF
--- a/core/models/tickets.go
+++ b/core/models/tickets.go
@@ -203,7 +203,7 @@ WHERE
 `
 
 // LookupTicketByUUID looks up the ticket with the passed in UUID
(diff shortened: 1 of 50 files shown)


rowanseymour (approved):

nicpottier:
Sorry Rowan, asked for your review a bit too quick, think it should be good to go now. I think this is way better, it actually lets us break the dependency from tasks->mailroom which is nice.

rowanseymour (approved):
I like it!

PR #437, opened 2021-06-08
🎟️ Ticket events

rowanseymour:
* Insert ticket opened events when new tickets are opened
* Insert closed and reopened events for tickets when endpoints are called
* Queue closed events for handling
* Add ticket closed triggers which can handle that


core/models/tickets.go
```diff
@@ -152,13 +164,12 @@ FROM
...
   t.org_id = $1 AND
-  t.id = ANY($2) AND
-  t.status = $3
+  t.id = ANY($2)
 `
 
 // LoadTickets loads all of the tickets with the given ids
-func LoadTickets(ctx context.Context, db Queryer, orgID OrgID, ids []TicketID, status TicketStatus) ([]*Ticket, error) {
-	return loadTickets(ctx, db, selectTicketsByIDSQL, orgID, pq.Array(ids), status)
+func LoadTickets(ctx context.Context, db Queryer, orgID OrgID, ids []TicketID) ([]*Ticket, error) {
```
rowanseymour: we already have 4 functions for loading tickets so making this one more generic

core/tasks/handler/worker.go
```diff
@@ -328,7 +349,7 @@ func HandleChannelEvent(ctx context.Context, db *sqlx.DB, rp *redis.Pool, eventT
 		trigger = models.FindMatchingMissedCallTrigger(oa)
 
 	case models.MOCallEventType:
-		trigger = models.FindMatchingIncomingCallTrigger(oa, modelContact)
+		trigger = models.FindMatchingIncomingCallTrigger(oa, contact)
```
rowanseymour, line 352: moving the code to load the flow contact up a bit means we can pass a flow contact here like we do for the other trigger types, and then in the trigger matching code we can have one unified way of getting the groups for a passed contact

core/models/tickets.go
```diff
@@ -86,6 +86,17 @@ func (t *Ticket) Config(key string) string {
 	return t.t.Config.GetString(key, "")
 }
 
+func (t *Ticket) FlowTicket(oa *OrgAssets) (*flows.TicketReference, error) {
+	ticketer := oa.TicketerByID(t.TicketerID())
+	if ticketer == nil {
+		return nil, errors.New("unable to load ticketer with id %d")
+	}
+
+	ticket := flows.NewTicketReference(t.UUID(), ticketer.Reference(), t.Subject(), t.Body(), string(t.ExternalID()))
```
rowanseymour: going back to goflow to see if I can get rid of this Ticket/TicketReference distinction...

core/models/users.go
```diff
@@ -0,0 +1,35 @@
...
+const (
+	// NilUserID is the id 0 considered as nil user id
+	NilUserID = UserID(0)
+)
+
+// UserID is our type for user ids, which can be null
+type UserID null.Int
+
+// MarshalJSON marshals into JSON. 0 values will become null
+func (i UserID) MarshalJSON() ([]byte, error) {
```
rowanseymour, line 18: @nicpottier I'm confused again about why these methods are needed.. why isn't being a `null.Int` enough? Also in general, would it maybe be better to have `UserID` and `NullableUserID` as two separate types?

(this is so that ticket events can be saved with created_on=NULL when they come from an external service)

web/ticket/testdata/close.json
```diff
@@ -13,6 +13,7 @@
         "path": "/mr/ticket/close",
         "body": {
             "org_id": 1,
+            "user_id": 3,
```
rowanseymour, line 16: latest RP now passes user_id to these endpoints

nicpottier (approved):

PR #457, opened 2021-07-08
😵 Add new error type for failed SQL queries

rowanseymour:
My cunning plan to get failing SQL queries into sentry didn't work very well because sentry trims the error message to 200 chars, so after all the wrapping you only get a few chars of SQL. This PR adds a new error type which can be used to attach SQL and args to the logrus entry as field data.


nicpottier (approved):
Huh, neat if it works!

PR #461, opened 2021-07-19
Log error if marking event fire as fired fails

rowanseymour:


core/tasks/campaigns/fire_campaign_event.go
```diff
@@ -93,7 +93,10 @@ func (t *FireCampaignEventTask) Perform(ctx context.Context, rt *runtime.Runtime
 	if len(contactMap) > 0 {
 		rc := rp.Get()
 		for _, failed := range contactMap {
-			marker.RemoveTask(rc, campaignsLock, fmt.Sprintf("%d", failed.FireID))
+			rerr := marker.RemoveTask(rc, campaignsLock, fmt.Sprintf("%d", failed.FireID))
+			if rerr != nil {
+				log.WithError(rerr).WithField("fire_id", failed.FireID).Error("error unmarking campaign fire")
+			}
```
rowanseymour, line 99: same as https://github.com/nyaruka/mailroom/blob/91fd4062ba437cf2f89a41438ed7e8a21519e03d/core/tasks/campaigns/fire_campaign_event.go#L61-L64

nicpottier (approved):
good to have but this is the opposite problem we saw ya? (event wasn't actually fired but was marked as fired)

rowanseymour:
@nicpottier well this is on the code that unmarks as fired, any contacts who failed.. i.e. any contacts not returned by `runner.FireCampaignEvents`.. so maybe?

PR #464, opened 2021-07-23
🧯 Update contact `modified_on` after populate dynamic group task

rowanseymour:


core/models/groups.go
```diff
@@ -409,5 +409,15 @@ func PopulateDynamicGroup(ctx context.Context, db *sqlx.DB, es *elastic.Client,
 		return 0, errors.Wrapf(err, "error marking dynamic group as ready")
 	}
 
+	// finally update modified_on for all affected contacts to ensure these changes are seen by rp-indexer
+	changed := make([]ContactID, 0, len(adds))
+	changed = append(changed, adds...)
+	changed = append(changed, removals...)
+
+	err = UpdateContactModifiedOn(ctx, db, changed)
```
nicpottier, line 417: Maybe this all belongs on AddContactsToGroups and DeleteContactsFromGroups instead?

nicpottier (approved):

core/models/groups.go
rowanseymour: I don't think so as those methods don't even hit the contacts_contact table and the other places we use them are when we're processing events/handlers/hooks and we've optimized updating modified_on for each contact.

nicpottier: Ok. Ya, just feels likely we are going to make this bug again elsewhere when we change contact group membership elsewhere by calling this method, but I suppose we can just try to be extra careful.

PR #466, opened 2021-07-26
🤦‍♂️ Actually save IVR messages with `sent_on` set

rowanseymour:
This is what happens when you don't add a unit test..


core/models/msgs.go
```diff
@@ -275,15 +275,15 @@ func NewOutgoingIVR(orgID OrgID, conn *ChannelConnection, out *flows.MsgOut, cre
...
-	m.SentOn = createdOn
+	m.SentOn = &createdOn
 	msg.SetChannelID(conn.ChannelID())
 
 	// if we have attachments, add them
 	for _, a := range out.Attachments() {
 		m.Attachments = append(m.Attachments, string(NormalizeAttachment(config.Mailroom, a)))
 	}
 
-	return msg, nil
```
rowanseymour, line 286: function can't actually error..

Issue #472, opened 2021-08-09
Support machine detection for IVR on Twilio 


norkans7:
https://www.twilio.com/docs/voice/answering-machine-detection#machinedetection

https://www.twilio.com/docs/voice/api/call-resource#create-a-call-resource

PR #473, opened 2021-08-09
Add support for Machine detection on Twilio calls

norkans7:
Addresses #472

### DIFF
--- a/core/ivr/twiml/twiml.go
+++ b/core/ivr/twiml/twiml.go
@@ -188,6 +188,9 @@ func (c *client) RequestCall(number urns.URN, callbackURL string, statusURL stri
 	form.Set("From", c.channel.Address())
 	form.Set("Url", callbackURL)
 	form.Set("StatusCallback", statusURL)
+	form.Set("MachineDetection", "Enabled")
+	form.Set("AsyncAMD", "true")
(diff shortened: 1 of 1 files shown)


core/ivr/twiml/twiml.go
```diff
@@ -286,6 +289,16 @@ func (c *client) ResumeForRequest(r *http.Request) (ivr.Resume, error) {
...
 // StatusForRequest returns the current call status for the passed in status (and optional duration if known)
 func (c *client) StatusForRequest(r *http.Request) (models.ConnectionStatus, int) {
+	answeredBy := r.Form.Get("AnsweredBy")
+
+	switch answeredBy {
+	case "human", "unknown", "":
+		break
+
+	default:
+		return models.ConnectionStatusErrored, 0
```
rowanseymour, line 299: `ConnectionStatusNoAnswer` ? and then let mailroom retry ?

rowanseymour: Ok so if it works for now to record as error then we can go with this, and possibly rework in future to use a different status.

norkans7: Ok

rowanseymour (approved):

PR #492, opened 2021-09-07
🙉 Add force param to close tickets endpoint to let us ignore errors

rowanseymour:
.. on an external ticket service when removing a ticketer

### DIFF
--- a/web/testing.go
+++ b/web/testing.go
@@ -65,7 +65,7 @@ func RunWebTests(t *testing.T, truthFile string, substitutions map[string]string
 
 		actualResponse []byte
 	}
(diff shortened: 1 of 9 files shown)


web/testing.go
```diff
@@ -65,16 +65,15 @@ func RunWebTests(t *testing.T, truthFile string, substitutions map[string]string
 
 		actualResponse []byte
 	}
-	tcs := make([]*TestCase, 0, 20)
+	tcs := make([]TestCase, 0, 20)
```
rowanseymour, line 68: we need `actual := tc` to copy by value.. the status check was never working

PR #520, opened 2021-11-09
Explicitly exclude msg fields from marshaling that courier doesn't use

rowanseymour:
Poking around at courier queues I noticed the message json has fields like `MsgType`, `ConnectionID` and `TopupID` which don't belong there.

```
	{
		"direction":"O",
		"modified_on":"2021-11-09T18:39:55.038607Z",
		"contact_urn_id":59726291,
		"session_id":970967356,
		"external_id":null,
		"MsgType":"F",
		"attachments":null,
                ...
```

### DIFF
--- a/core/models/msgs_test.go
+++ b/core/models/msgs_test.go
@@ -1,15 +1,18 @@
 package models_test
 
 import (
+	"encoding/json"
 	"fmt"
 	"testing"
 	"time"
 
 	"github.com/nyaruka/gocommon/dates"
+	"github.com/nyaruka/gocommon/jsonx"
 	"github.com/nyaruka/gocommon/urns"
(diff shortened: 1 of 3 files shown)


core/models/msgs_test.go
```diff
@@ -22,6 +25,11 @@ import (
 func TestOutgoingMsgs(t *testing.T) {
 	ctx, rt, db, _ := testsuite.Get()
 
+	defer func() {
+		testsuite.Reset(testsuite.ResetData)
+		db.MustExec(`UPDATE orgs_org SET is_suspended = FALSE`)
```
rowanseymour, line 30: this test leaks changes to this in subsequent tests

PR #535, opened 2021-11-29
Update sessions and runs in batches when exiting

rowanseymour:


norkans7 (commented):
I think we need a similar strategy for 

https://github.com/nyaruka/mailroom/blob/7f937bb8ae305fd05381044ab7835b3b31d0b036/core/models/runs.go#L1108

rowanseymour:
`InterruptContactRuns` is only called when starting flows - a thing which is already batched by contacts

PR #538, opened 2021-12-02
Rework how we create outgoing messages, and fix retries of messages with diff priorities

rowanseymour:


core/models/msgs.go
```diff
@@ -446,6 +453,7 @@ SELECT
 	msg_count,
 	error_count,
 	next_attempt,
+	coalesce(high_priority, FALSE) as high_priority,
```
rowanseymour, line 456: because incoming messages have this set to `NULL` but struct just has a `bool`

core/msgio/courier.go
```diff
@@ -98,19 +99,16 @@ func QueueCourierMessages(rc redis.Conn, contactID models.ContactID, msgs []*mod
 		assert(msg.Channel().Type() != models.ChannelTypeAndroid, "can't queue an android message to courier")
 		assert(msg.URN() != urns.NilURN && msg.ContactURNID() != nil, "can't queue a message to courier without a URN")
 
-		// same channel? add to batch
-		if msg.Channel() == currentChannel {
+		// if this msg is the same channel and priority, add to current batch, otherwise start new batch
+		if msg.Channel() == currentChannel && msg.HighPriority() == currentPriority {
```
rowanseymour, line 103: this fixes where we might be retrying multiple messages for the same contact with different priorities.. not entirely sure that's a situation that arises with any frequency

Issue #547, opened 2021-12-08
Create failed messages where we don't have a channel or URN


rowanseymour:
And set `failed_reason` accordingly. See https://github.com/nyaruka/mailroom/blob/main/core/handlers/msg_created.go#L74-L76

PR #551, opened 2021-12-13
Write failed messages when missing channel or URNs

norkans7:
Addresses https://github.com/nyaruka/mailroom/issues/547


core/handlers/msg_created.go
```diff
@@ -71,20 +70,11 @@ func handleMsgCreated(ctx context.Context, rt *runtime.Runtime, tx *sqlx.Tx, oa
...
-
 	// messages in messaging flows must have urn id set on them, if not, go look it up
 	if scene.Session().SessionType() == models.FlowTypeMessaging {
 		urn := event.Msg.URN()
 		if models.GetURNInt(urn, "id") == 0 {
-			urn, err := models.GetOrCreateURN(ctx, tx, oa, scene.ContactID(), event.Msg.URN())
-			if err != nil {
-				return errors.Wrapf(err, "unable to get or create URN: %s", event.Msg.URN())
-			}
+			urn, _ := models.GetOrCreateURN(ctx, tx, oa, scene.ContactID(), event.Msg.URN())
```
rowanseymour: we still want to err if this fails.. I think you want a check for `event.Msg.URN() != urns.NilURN` above

core/handlers/msg_created.go
```diff
@@ -94,9 +84,6 @@ func handleMsgCreated(ctx context.Context, rt *runtime.Runtime, tx *sqlx.Tx, oa
 	var channel *models.Channel
 	if event.Msg.Channel() != nil {
 		channel = oa.ChannelByUUID(event.Msg.Channel().UUID)
```
rowanseymour: likewise.. it's still an error for the engine to pass back an invalid channel

core/models/msgs.go
```diff
@@ -83,6 +83,8 @@ const (
 	MsgFailedLooping    = MsgFailedReason("L")
 	MsgFailedErrorLimit = MsgFailedReason("E")
 	MsgFailedTooOld     = MsgFailedReason("O")
+	MsgFailedChannel    = MsgFailedReason("C")
+	MsgFailedURN        = MsgFailedReason("U")
```
rowanseymour: We can merge these since the engine looks for compatible pairs of channel+URN and either returns a message with both set or neither set.. maybe `MsgFailedDestination` ? That's what the engine calls a channel+URN pair

core/models/msgs.go
```diff
@@ -385,7 +387,17 @@ func newOutgoingMsg(rt *runtime.Runtime, org *Org, channel *Channel, contactID C
 	msg.SetChannel(channel)
 
 	if err := msg.SetURN(out.URN()); err != nil {
-		return nil, errors.Wrapf(err, "error setting msg URN")
+		msg.SetURN(urns.NilURN)
```
rowanseymour: again I think this should be `if msg.URN then SetURN`

rowanseymour (changes requested):

core/handlers/msg_created.go
norkans7: when channel is `nil` the event will be ignored and not handled here?
I thought we need to ignore that error here so the message is saved with the Failed status

rowanseymour (approved):

PR #557, opened 2022-01-04
Clamp flow expiration values to valid ranges when loading flows

rowanseymour:
We got a lot of flows with junk values which we can also clean up on the RP side but important that engine always sets valid expirations

### DIFF
--- a/core/models/flows.go
+++ b/core/models/flows.go
@@ -179,8 +179,13 @@ SELECT ROW_TO_JSON(r) FROM (SELECT
 		jsonb_build_object(
 			'name', f.name,
 			'uuid', f.uuid,
-			'flow_type', f.flow_type, 
-			'expire_after_minutes', f.expires_after_minutes,
+			'flow_type', f.flow_type,
+			'expire_after_minutes', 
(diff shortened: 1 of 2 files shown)


core/models/flows.go
```diff
@@ -179,8 +179,13 @@ SELECT ROW_TO_JSON(r) FROM (SELECT
...
 			'uuid', f.uuid,
-			'flow_type', f.flow_type, 
-			'expire_after_minutes', f.expires_after_minutes,
+			'flow_type', f.flow_type,
+			'expire_after_minutes', 
+				CASE f.flow_type 
+				WHEN 'M' THEN GREATEST(5, LEAST(f.expires_after_minutes, 43200))
+				WHEN 'V' THEN GREATEST(1, LEAST(f.expires_after_minutes, 15))
+				ELSE 0
+				END,
```
rowanseymour, line 188: see https://github.com/nyaruka/rapidpro/blob/26cf16974aa19416d1cb89ee34430668ae86f9c6/temba/flows/models.py#L128-L155

norkans7 (approved):

PR #559, opened 2022-01-05
Start writing FlowSession.wait_expires_on

rowanseymour:


core/models/runs.go
```diff
@@ -535,18 +537,21 @@ func (s *Session) FlowSession(cfg *runtime.Config, sa flows.SessionAssets, env e
...
 	s.s.WaitStartedOn = nil
+	s.s.WaitExpiresOn = nil
 	s.s.TimeoutOn = nil
 	s.timeout = nil
 
 	for _, e := range evts {
 		switch typed := e.(type) {
 		case *events.MsgWaitEvent:
+			now := time.Now()
+			s.s.WaitStartedOn = &now
```
rowanseymour, line 548: we want this set regardless of whether there is a timeout on the wait so that we can adjust `WaitExpiresOn` if a user edits the expiry on the flow

core/models/runs.go
```diff
@@ -596,6 +601,7 @@ func (s *Session) Update(ctx context.Context, rt *runtime.Runtime, tx *sqlx.Tx,
 	// set our sprint, wait and step finder
 	s.sprint = sprint
 	s.findStep = fs.FindStep
+	s.s.CurrentFlowID = NilFlowID
```
rowanseymour, line 604: if the session has exited, we want this to be cleared

PR #560, opened 2022-01-05
Correctly set ended_on on session objects and add tests

rowanseymour:
And fix not applying events from failed sessions


core/models/runs.go
```diff
@@ -865,7 +875,7 @@ func WriteSessions(ctx context.Context, rt *runtime.Runtime, tx *sqlx.Tx, org *O
 	// apply our all events for the session
 	scenes := make([]*Scene, 0, len(ss))
 	for i := range sessions {
-		if ss[i].Status() == SessionStatusFailed {
+		if ss[i].Status() == flows.SessionStatusFailed {
```
rowanseymour, line 878: `"failed" != "F"` this was a bug not caught because `SessionStatusFailed` was untyped

core/models/runs.go
```diff
@@ -38,11 +38,11 @@ type FlowRunID int64
 const NilFlowRunID = FlowRunID(0)
 
 const (
-	SessionStatusWaiting     = "W"
-	SessionStatusCompleted   = "C"
-	SessionStatusExpired     = "X"
-	SessionStatusInterrupted = "I"
-	SessionStatusFailed      = "F"
+	SessionStatusWaiting     SessionStatus = "W"
```
rowanseymour, line 41: these were untyped strings

PR #578, opened 2022-01-26
we should not try to import rows without UUID, URNs or Name at least

norkans7:


core/models/imports.go
```diff
@@ -204,6 +204,12 @@ func (b *ContactImportBatch) getOrCreateContacts(ctx context.Context, db Queryer
 		spec := imp.spec
 
 		uuid := spec.UUID
+
+		if uuid == "" && spec.Name == nil && len(spec.URNs) == 0 {
```
rowanseymour: 1. Doesn't seem like we should allow contact rows with just a name. Re-importing will always create new contacts
2. Shouldn't RapidPro be stripping rows like these out before we even get to mailroom?

norkans7: Adjusted to not allow rows with just a name, and added the changes to rapidpro side too https://github.com/nyaruka/rapidpro/pull/3666

rowanseymour:
Closing in favor of https://github.com/nyaruka/rapidpro/pull/3666 which stops RP from creating specs like these in the first place

PR #590, opened 2022-02-07
When interrupting sessions for flow starts, ignore flow type

rowanseymour:


core/models/sessions.go
```diff
@@ -807,21 +807,10 @@ func WriteSessionOutputsToStorage(ctx context.Context, rt *runtime.Runtime, sess
 	return nil
 }
 
-const sqlSelectWaitingSessionOverlap = `
-SELECT DISTINCT(contact_id)
-  FROM flows_flowsession fs 
-  JOIN flows_flow ff ON fs.current_flow_id = ff.id
- WHERE fs.status = 'W' AND ff.is_active = TRUE AND ff.is_archived = FALSE AND ff.flow_type = $1 AND fs.contact_id = ANY($2)`
```
rowanseymour, line 814: removing the check on the flow here... when a flow is deleted / archived we interrupt waiting sessions and that process correctly takes care of clearing `contact.current_flow_id`

PR #592, opened 2022-02-14
Add support for querying by flow

rowanseymour:


core/models/flows.go
```diff
@@ -165,111 +170,56 @@ func loadFlow(ctx context.Context, db Queryer, sql string, orgID OrgID, arg inte
...
+					'uuid', f.uuid, 
+					'id', f.id,
+					'name', f.name,
+					'revision', revision, 
+					'expires', f.expires_after_minutes
+				)
+		) as definition
+	FROM
+		flows_flow f
+	INNER JOIN LATERAL (
```
rowanseymour, line 205: my first LATERAL join - allows us to reference `f.id` from the outer select instead of re-querying on the flow id/name/uuid

norkans7 (approved):

PR #596, opened 2022-03-08
Don't try to resume expired session if contact isn't active

rowanseymour:


core/tasks/expirations/cron.go
```diff
@@ -77,7 +77,7 @@ func HandleWaitExpirations(ctx context.Context, rt *runtime.Runtime) error {
 		}
 
 		// if it can't be resumed, add to batch to be expired
-		if !expiredWait.CanResume {
+		if !expiredWait.WaitResumes || expiredWait.ContactStatus != models.ContactStatusActive {
```
rowanseymour, line 80: I think long term we want to rethink this an actually allow non-active contacts in flows - just make it so that they can't be messaged.

PR #603, opened 2022-03-16
Wait for crons on shutdown

rowanseymour:
Also simplifies the starting of cron jobs

### DIFF
--- a/cmd/mailroom/main.go
+++ b/cmd/mailroom/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"os"
 	"os/signal"
 	goruntime "runtime"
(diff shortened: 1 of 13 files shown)


cmd/mailroom/main.go
```diff
@@ -108,11 +109,12 @@ func main() {
 
 // handleSignals takes care of trapping quit, interrupt or terminate signals and doing the right thing
 func handleSignals(mr *mailroom.Mailroom) {
-	sigs := make(chan os.Signal)
+	sigs := make(chan os.Signal, 1)
```
rowanseymour, line 111: gets rid of linter warning on line below that you should us a buffered channel with `signal.Notify`

PR #612, opened 2022-04-04
Fix unstopping of contacts who message in

### LINKED ISSUE #611: Stopped contacts that message us are unstopped but started in the flow as stopped
We do unstop them, but we've already loaded the flow contact and so they're started as a stopped contact in the flow.

rowanseymour:
Fixes #611


rowanseymour:
cc @tybritten

PR #644, opened 2022-07-25
Failed sessions should be inserted with `ended_on` set

rowanseymour:
We set this correctly if things fail on later resumes, but in the insert case after the first sprint, we only set ended_on for status=C sessions.


core/models/sessions.go
```diff
@@ -640,26 +637,25 @@ func InsertSessions(ctx context.Context, rt *runtime.Runtime, tx *sqlx.Tx, oa *O
...
+	insertEndedSQL := sqlInsertEndedSession
+	insertWaitingSQL := sqlInsertWaitingSession
 
 	// if writing our sessions to S3, do so
 	if rt.Config.SessionStorage == "s3" {
 		err := WriteSessionOutputsToStorage(ctx, rt, sessions)
 		if err != nil {
-			// for now, continue on for errors, we are still reading from the DB
-			logrus.WithError(err).Error("error writing sessions to s3")
+			return nil, errors.Wrapf(err, "error writing sessions to storage")
```
rowanseymour, line 648: this is a holdover from when we always read output from DB and we were just testing S3 writes - if S3 writing fails now, that's an error

core/models/sessions_test.go
```diff
@@ -253,6 +249,51 @@ func TestSessionWithSubflows(t *testing.T) {
...
+	testFlows := testdata.ImportFlows(db, testdata.Org1, "testdata/ping_pong.json")
+	ping := testFlows[0]
+
+	oa, err := models.GetOrgAssetsWithRefresh(ctx, rt, testdata.Org1.ID, models.RefreshFlows)
+	require.NoError(t, err)
+
+	modelContact, _ := testdata.Cathy.Load(db, oa)
+
+	_, flowSession, sprint1 := test.NewSessionBuilder().WithAssets(oa.SessionAssets()).WithFlow(ping.UUID).
+		WithContact(testdata.Bob.UUID, flows.ContactID(testdata.Cathy.ID), "Cathy", "eng", "").MustBuild()
```
norkans7: typo `testdata.Bob.UUID`

PR #647, opened 2022-08-02
Allow Mailroom to use AWS Credential Chains for accessing S3

tybritten:
If a user wants to enable s3 access, but doesn't want to use static key/secret values, they can set `AWSUseCredChain` to true, and the AWS SDK will use its built-in credential chain (env vars, file, role, etc) to access S3.


rowanseymour (commented):
Looks fine but seems you don't have gofmt on save setup?

rowanseymour (approved):

PR #662, opened 2022-09-20
Add endpoint to fail messages

norkans7:


rowanseymour:
Not sure about this.. mailroom marking messages as failed isn't inherently more performant than RapidPro doing it.. you'd still need to do it batches asynchrously.

Can't we have courier just fail them? It's going to end up processing them anyway (assuming we don't clear out the queues)

norkans7:
Ohh yes, we need to use the batches, this is going to be used by the channel release

rowanseymour:
Ok so supposing you batch the messages on the RP side.. `Msg.objects.filter(id__in=batch_ids).update(status='F')` isn't going be any slower than calling a mailroom endpoint to do exactly the same thing (probably be faster).

Now obviously performance isn't the only concern - it's nice to have one place that knows how to do one thing, but knowing how to fail a message isn't particularly complex.

What about courier doing it? That avoids any concerns about race conditions between RP and courier which is already going to be processing these messages.

norkans7:
Already there were messages queued on courier to be processed for a channel that was deleted and those are failing to be processed https://sentry.io/organizations/nyaruka/issues/3552391395/

I added https://github.com/nyaruka/courier/pull/514 to make sure those are failed by courier

norkans7:
we can also simply do the batches on https://github.com/nyaruka/rapidpro/blob/0b501e16ee45ac0038565f21cd137cf3bb937191/temba/channels/models.py#L924 instead of these changes

rowanseymour:
Replaced by https://github.com/nyaruka/mailroom/pull/670

PR #670, opened 2022-10-04
Interrupt channel

norkans7:


core/models/msgs.go
```diff
@@ -1273,6 +1312,38 @@ func ResendMessages(ctx context.Context, db Queryer, rp *redis.Pool, oa *OrgAsse
 	return resent, nil
 }
 
+const sqlUpdateMsgStatusFailed = `
+UPDATE msgs_msg m
+   SET status = 'F', modified_on = NOW()
+ WHERE id = ANY($1)`
+
+// FailMessages prepares messages for failing and marking them as FAILED
+func FailMessages(ctx context.Context, db Queryer, rp *redis.Pool, oa *OrgAssets, msgs []*Msg) ([]*Msg, error) {
```
rowanseymour: Why do we need to load the messages? I think you can just directly update messages with status = Q,P,E for a given channel. We can still do it batches like...

```
WITH rows AS (
  SELECT id FROM msgs_msg WHERE ... LIMIT 1000
)
UPDATE msgs_msg SET status = 'F'
WHERE id IN (SELECT id FROM rows)
```

And just keep calling that until it updates zero rows

core/msgio/courier.go
```diff
@@ -116,3 +116,24 @@ func QueueCourierMessages(rc redis.Conn, contactID models.ContactID, msgs []*mod
...
+var clearChannelQueueScript = redis.NewScript(3, `
+-- KEYS: [QueueType, QueueName, TPS]
+local queueType, queueName, tps = KEYS[1], KEYS[2], tonumber(KEYS[3])
+
+-- first construct the base key for this queue from the type + name + tps, e.g. "msgs:0a77a158-1dcb-4c06-9aee-e15bdf64653e|10"
+local queueKey = queueType .. ":" .. queueName .. "|" .. tps
+
+-- clear the sorted sets for the key
+redis.call("ZREMRANGEBYSCORE", queueKey .. "/1", "-inf", "inf")
+redis.call("ZREMRANGEBYSCORE", queueKey .. "/0", "-inf", "inf")
```
rowanseymour: can't we just `DEL` both sets ?

rowanseymour (changes requested):

core/models/msgs.go
norkans7: Updated

core/tasks/channels/interrupt_channel.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Perform implements tasks.Task
+func (t *InterruptChannelTask) Perform(ctx context.Context, rt *runtime.Runtime, orgID models.OrgID) error {
+	db := rt.DB
+	rc := rt.RP.Get()
+	defer rc.Close()
+
+	channelIDs := []models.ChannelID{t.ChannelID}
+
+	channels, err := models.GetChannelsByID(ctx, db, channelIDs)
```
rowanseymour, line 34: after we switch over RP to use this task rather than `interrupt_sessions`, we can replace this with `models.GetChannelByID(ctx, db, channelID)`

core/tasks/channels/interrupt_channel.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return err
+	}
+
+	channel := channels[0]
+
+	if err := models.InterruptSessionsForChannels(ctx, db, channelIDs); err != nil {
+		return err
```
rowanseymour: use `errors.Wrap` and add context

rowanseymour (changes requested):

rowanseymour:
@norkans7 just wrap some of those errors and I think this is looking good

rowanseymour (approved):

Issue #691, opened 2022-12-01
Not retrying of PENDING outgoing messages


rowanseymour:
If we fail to queue outgoing messages, we put them into a pending state. But then nothing happens to them until they're failed a week later. Doesn't happen very often but it's possible when redis fails.

PR #692, opened 2022-12-05
Retry messages which fail to queue to courier

rowanseymour:
Addresses https://github.com/nyaruka/mailroom/issues/691, requires https://github.com/nyaruka/rapidpro/pull/4179


ericnewcomer (approved):
How do messages break out of this retry cycle?

rowanseymour:
They don't but I think that's ok because this isn't expected behavior like a channel not working - this is something broken in our deployment

PR #703, opened 2026-03-05
Add secondary URN support to msg_received task

norkans7:
## Summary
Add optional `SecondaryURN` field to the `MsgReceived` task. When provided, the secondary URN is appended to the contact if not already present, enabling channels to pass additional identifiers that should be tracked alongside the primary URN.

## Changes
- Added `SecondaryURN` field to `MsgReceived` struct (JSON: `secondary_urn`, omitempty)
- Apply URN append modifier after affinity modifier in `perform()` to add the secondary URN
- Added test case verifying secondary URN is added and deduplication works correctly

## Test Plan
- New `TestMsgReceivedSecondaryURN` test verifies secondary URN addition and handles duplicates
- Existing tests pass without modification (field is optional)


norkans7:
wrong repo

